PR #184, opened 2015-01-14
Add and test the second kind of hotness followup message.

ralphbean:
This handles the new message type added in
fedora-infra/the-new-hotness#9.


pypingou:
Cool :+1:

PR #185, opened 2015-01-14
No longer stuff package owners in msg2usernames for pkgdb.

ralphbean:
It used to be that, for pkgdb messages, msg2usernames included many
usernames:  the actor performing the event, the subject on whom the action
was performed, and any package owners of the package.

That third bit was added as a hack to get message routing in FMN working
easily when it first launched.  I wanted to route pkgdb messages to package
owners when requests were made on their packages.

This is no longer necessary since FMN can now query pkgdb for who owns what
packages.. but furthermore it actually breaks other more complicated
filters that we're adding to the default set.

For instance, there is a rule that says "if someone does something to one
of my packages, tell me unless I did it".  The "unless I did it" part is
artificially evaluating to true, since everyone's usernames get returned by
msg2usernames.  This commit should fix that.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/tests/pkgdb.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/pkgdb.py
@@ -366,7 +366,7 @@ class LegacyTestPkgdbBranchClone(Base):
     }
 
 
-class TestPkgdbCritpathUpdate(Base):
+class TestLegacyPkgdbCritpathUpdate(Base):
     """ The Fedora `Package DB <https://admin.fedoraproject.org/pkgdb>`_
     publishes messages on this topic when the critical path status of a
     package changes (when it is either added, or removed from the critical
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/tests/pkgdb.py
```diff
@@ -366,7 +366,7 @@ class LegacyTestPkgdbBranchClone(Base):
     }
 
 
-class TestPkgdbCritpathUpdate(Base):
+class TestLegacyPkgdbCritpathUpdate(Base):
```
pypingou: There is still a pkgdb2.package.critpatch.update message: https://github.com/fedora-infra/pkgdb2/blob/master/pkgdb2/lib/utils.py#L264

pypingou:
:+1: on the core of the change and one remark for the passage to Legacy of one test

fedmsg_meta_fedora_infrastructure/tests/pkgdb.py
ralphbean: Yes.  That one is tested here:  https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure/blob/feature/pkgdb-usernames-kludge-removal/fedmsg_meta_fedora_infrastructure/tests/pkgdb.py#L1483-L1499

PR #191, opened 2015-01-17
Handle hotness.project.map messages.

ralphbean:
These new messages get introduced in fedora-infra/the-new-hotness#10.


pypingou:
LGTM :+1:

PR #196, opened 2015-01-28
Handle new sigul messages (via koji).

ralphbean:


fedmsg_meta_fedora_infrastructure/buildsys.py
```diff
@@ -101,8 +105,6 @@ def subtitle(self, msg, **config):
                 tmpl = tmpl[len("{owner}'s "):]
 
             return tmpl.format(inst=inst, **msg['msg'])
-        else:
```
puiterwijk: Why remove the NotImplementedError?

fedmsg_meta_fedora_infrastructure/buildsys.py
```diff
@@ -159,14 +161,16 @@ def packages(self, msg, **config):
...
             return set([msg['msg']['package']])
+        elif 'buildsys.rpm.sign' in msg['topic']:
+            return set([msg['msg']['info']['name']])
         elif 'buildsys.build.state.change' in msg['topic']:
             return set([msg['msg']['name']])
         elif 'buildsys.task.state.change' in msg['topic']:
             # We can't *really* associate scratch builds with a package,
             # honestly.
             return set([])
-        else:
```
puiterwijk: You should probably keep this in

fedmsg_meta_fedora_infrastructure/buildsys.py
```diff
@@ -274,5 +281,7 @@ def objects(self, msg, **config):
                 'tags',
                 msg['msg'].get('tag', 'unknown'),
             ])])
-        else:
-            raise NotImplementedError("%r" % msg)
+        elif 'buildsys.rpm.sign' in msg['topic']:
+            return set(['signatures/' + msg['msg']['info']['name']])
+
```
puiterwijk: You should probably retain a else:

fedmsg_meta_fedora_infrastructure/buildsys.py
```diff
@@ -60,6 +60,10 @@ def subtitle(self, msg, **config):
             tmpl = self._(
                 "Package list change for {package}:  '{tag}'{inst}")
             return tmpl.format(inst=inst, **msg['msg'])
+        elif 'buildsys.rpm.sign' in msg['topic']:
+            tmpl = self._('Koji imported a gpg signature for '
```
puiterwijk: Is this a good summary?
I'd say "Koji package signed: %s" or the like.

fedmsg_meta_fedora_infrastructure/buildsys.py
ralphbean: Because it spams us with errors when a new message gets introduced instead of just returning an empty string.

I didn't even realize these errors were happening until pingou noticed his inbox getting full.  With this change, the messages would still pass through to IRC, but as obviously new messages that need to be dealt with.

fedmsg_meta_fedora_infrastructure/buildsys.py
ralphbean: No, I'd rather return ``set()`` for the reasons mentioned above.

fedmsg_meta_fedora_infrastructure/buildsys.py
ralphbean: Ah, but koji didn't sign the package -- someone else did (this is usually sigul in our case).

fedmsg_meta_fedora_infrastructure/buildsys.py
puiterwijk: Okay, makes sense.

fedmsg_meta_fedora_infrastructure/buildsys.py
puiterwijk: Right, but then I guess it should be:
else:
    return set()

fedmsg_meta_fedora_infrastructure/buildsys.py
puiterwijk: "Koji package signed" doesn't say koji signed the package, just that a package built by koji is signed :).
I think more people might agree with me that "imported a gpg signature" makes me think of importing a signature for a new key.

fedmsg_meta_fedora_infrastructure/buildsys.py
ralphbean: But.. there's a ``return set()`` at the bottom?

Either way, if it returns ``None`` (like it would if there's no return statement) there is logic in the code that calls this function that knows that ``set()`` should be the default value if no explicit value is returned.

puiterwijk: Right, so the return set is there, but no else. I guess since the rest is also return's, it's okay.

puiterwijk:
Other than that I find the title a bit counter-intuitive, :+1:

ralphbean:
Thanks @puiterwijk!  As mentioned in IRC:  yeah, I don't like the alternative either, though.  that's easy enough to change later if we find a text we both like.

fedmsg_meta_fedora_infrastructure/buildsys.py
pypingou: ``Package XX was signed in koji`` or maybe ``Package XX in koji has been signed`` ?

PR #197, opened 2015-01-30
Mention the sigkey in the rpm.sign subtitle.

ralphbean:


pypingou:
This is definitely a better solution than what we had, :+1: :fireworks:

PR #199, opened 2015-02-10
Demote this error message.

ralphbean:
We get emails every time log.error is called.  The problem with this one
is that it happens quite often, and it's not really actionable.  There's
nothing for us to do about it.  Better to reduce email noise so we can
actually respond to other errors that are more significant.


pypingou:
:+1:  and thanks :)

PR #206, opened 2015-02-18
Convert github.watch messages to github.star.

ralphbean:
It turns out that github's event system is weird and their
``X-Github-Event: watch`` header actually means that someone starred a
repository.

This updates us to make more sense in our links and text.

An update to github2fedmsg that converts the messages will be coming
soon.


pypingou:
LGTM :+1:

Issue #54, opened 2014-01-14
bodhi avatars seem broken


ralphbean:
Take a look at this:

https://apps.fedoraproject.org/datagrepper/raw/?package=hibernate&order=desc&rows_per_page=100

It shows the correct avatar (secondary icon) for git messages, but not for bodhi.

Is that supposed to be the package icon?

PR #210, opened 2015-02-23
Implement a long-form value for koji builds.

ralphbean:
This should fix fedora-infra/fmn#54.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/buildsys.py
+++ b/fedmsg_meta_fedora_infrastructure/buildsys.py
@@ -21,6 +21,30 @@ from fedmsg_meta_fedora_infrastructure import BaseProcessor
 
 from fasshim import gravatar_url
 
+import datetime
+from pytz import UTC
+
+try:
+    import koji
+except ImportError:
+    pass
+
+
+_build_template = """Package:    {name}-{version}-{release}
+Status:     {status}
+Built by:   {owner_name}
+ID:         {id}
+Started:    {started}
+Finished:   {finished}
+
+Closed tasks:
+-------------
+"""
+
+_task_header_template = """Task {id} on {build_host}
+Task Type: {method} ({arch})
+"""
+
 
 class KojiProcessor(BaseProcessor):
     __name__ = "buildsys"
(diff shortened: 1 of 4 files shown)


fedmsg_meta_fedora_infrastructure/buildsys.py
```diff
@@ -31,6 +55,67 @@ class KojiProcessor(BaseProcessor):
...
+
+        children = session.getTaskChildren(taskid)
+        for child in sorted(children, key=lambda d: d['completion_ts']):
+            retval += "\n" + cls._fill_task_template(session, child['id'])
+        return retval
+
+    @classmethod
+    def _fill_build_template(cls, session, build):
+        full_build = session.getBuild(build['build_id'])
+        lookup = dict(zip(*zip(*koji.BUILD_STATES.items())[::-1]))
```
pypingou: Won't this fail if the import of koji failed?

fedmsg_meta_fedora_infrastructure/buildsys.py
```diff
@@ -31,6 +55,67 @@ class KojiProcessor(BaseProcessor):
...
+            url = "https://koji.fedoraproject.org/kojihub"
+        elif instance == 'ppc':
+            url = "http://ppc.koji.fedoraproject.org/kojihub"
+        elif instance == 's390':
+            url = "http://s390.koji.fedoraproject.org/kojihub"
+        elif instance == 'arm':
+            url = "http://arm.koji.fedoraproject.org/kojihub"
+
+        if 'buildsys.build.state.change' in msg['topic']:
+            session = koji.ClientSession(url)
```
pypingou: Same as above :)

fedmsg_meta_fedora_infrastructure/tests/buildsys.py
```diff
@@ -310,7 +348,7 @@ class TestKojiBuildStateChangeStartNoOwner(Base):
     }
 
 
-class TestKojiBuildStateChangeFail(Base):
```
pypingou: No test for fail then?

fedmsg_meta_fedora_infrastructure/buildsys.py
ralphbean: Yes.  :)

fedmsg_meta_fedora_infrastructure/tests/buildsys.py
ralphbean: Not anymore... before there was a test for 'start' and 'fail', but none for 'cancel' or 'complete'.  Now there's a test for 'start' and 'complete', but none for 'cancel' or 'fail'.

pypingou: Would it make sense to include 'cancel' or 'fail' as they are likely going to have different info compared to 'complete' (I'm thinking especially about the long_form).

Note: I won't block the PR on this, it can always be added later if we wish to :)

ralphbean: No, you're right.  We should handle that.  Thanks

pypingou:
Cool, many thanks sir! :+1:

pypingou:
Still :+1: for me

PR #212, opened 2015-02-24
(anitya) Substitute Fedora package name for project name if available.

ralphbean:
This should fix fedora-infra/the-new-hotness#21.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/anitya.py
+++ b/fedmsg_meta_fedora_infrastructure/anitya.py
@@ -128,7 +128,7 @@ class AnityaProcessor(BaseProcessor):
             # substitute "our" package name for the project name to make
             # everyone's emails more readable.
             # https://github.com/fedora-infra/the-new-hotness/issues/21
-            for package in message['packages']:
+            for package in message.get('packages', []):
                 if package['distro'] == 'Fedora':
                     project = package['package_name']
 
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/anitya.py
```diff
@@ -118,10 +118,20 @@ def subtitle(self, msg, **config):
...
             if 'message' in msg['msg']:
                 message = msg['msg']['message']
             else:
                 message = msg['msg']
+
+            # If this project is mapped to a Fedora package, then just
+            # substitute "our" package name for the project name to make
+            # everyone's emails more readable.
+            # https://github.com/fedora-infra/the-new-hotness/issues/21
+            for package in message['packages']:
```
pypingou: Should we use ``.get()``?

pypingou:
One question, otherwise :+1:

fedmsg_meta_fedora_infrastructure/anitya.py
ralphbean: Heh, what does anitya return?  I thought it always had at least a ``[]`` for 'packages'.  Might as well be on the safe side.

pypingou: That would be my assumption as well, but I didn't check the code and safe_side++ :)

PR #213, opened 2015-02-24
Make github longform tests conditional.

ralphbean:
Also, move zodbot tests to their own file, and make the long_form part
conditional like the others.


pypingou:
:+1: for me, looks like in the long term there might be more tests we could move out of ``__init__.py`` :smile:

ralphbean:
> there might be more tests we could move out of __init__.py

Agreed.  I'm doing it slowly as I run into a need to change them.. ;P

PR #216, opened 2015-03-06
There is more info here now...

ralphbean:


pypingou:
Let's wait for travis, but otherwise :+1: for me

pypingou:
And travis agrees :)

PR #221, opened 2015-03-08
Use FAS openid libravatar first for git.receive messages

Ghost-script:
Added an example of modern git.receive message and a handler for it.


ralphbean:
Ah, there are some other ``git.receive`` tests hidden in ``fedmsg_meta_fedora_infrastructure/tests/__init__.py``.  Like this one:  https://github.com/Ghost-script/fedmsg_meta_fedora_infrastructure/blob/openid_libravatar/fedmsg_meta_fedora_infrastructure/tests/__init__.py#L532-L534

See at the very bottom here that they make the test suite fail:  https://travis-ci.org/fedora-infra/fedmsg_meta_fedora_infrastructure/jobs/53566780

Can you update those tests to have the new correct ``expected_secondary_icon``?

Ghost-script:
@ralphbean i actually copied the same test for scm new style message in this patch. I think there is no point in running same tests twice. So i will remove the new test i added to `tests/scm.py` and update the tests in `__init__ .py` . What do you think about this?

ralphbean:
> Or we can move the test from __init__.py to tests/scm.py

You could move them all from ``__init__.py`` to ``tests/scm.py``.  That would be helpful.  :)

At one point in the past they were all in the same __init__.py file, but I've been trying to slowly move them out to their own files for readability.  This would help that.

fedmsg_meta_fedora_infrastructure/scm.py
```diff
@@ -33,6 +33,9 @@ class SCMProcessor(BaseProcessor):
 
     def secondary_icon(self, msg, **config):
         if '.git.receive' in msg['topic']:
+            if 'username' in msg['msg']['commit']:
+                username = msg['msg']['commit']['username']
```
Ghost-script: Just want to confirm, is `msg['msg']['commit']['username']`  the fas_username? or github username? we need fas_username to generate the libravatar url..

ralphbean: In ``scm.py`` here it should just be the ``fas_username``.

Ghost-script:
@ralphbean, not sure but the `conglomerate/bodhi/test_comment.py` gives assertion error. I fpasted the output.
http://fpaste.org/196113/25969493/

ralphbean:
I... don't know why it's failing either.  I couldn't duplicate it when I pulled.

What version of ``fedmsg`` do you have installed?

Also, could you try making this change to see if it makes a difference?

```diff
diff --git a/fedmsg_meta_fedora_infrastructure/tests/scm.py b/fedmsg_meta_fedora_infrastructure/tests/scm.py
index e46750d..d7a53d8 100644
--- a/fedmsg_meta_fedora_infrastructure/tests/scm.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/scm.py
@@ -23,7 +23,7 @@
 import os
 import unittest
 
-from fedmsg.tests.test_meta import Base
+from fedmsg_meta_fedora_infrastructure.tests.base import Base
 
 from common import add_doc
 ```

**Comment by Ghost-script** [comment]
@ralphbean , changed that import, same output. How to check fedmsg version?

**Comment by ralphbean** [comment]
> How to check fedmsg version?

In the virtualenv, try ``pip freeze | grep fedmsg``.  I get:

```
fedmsg==0.12.2
-e git+git@github.com:fedora-infra/fedmsg_meta_fedora_infrastructure.git@45c2f47d14a610b4e736eadb8fedac91e7ed148a#egg=fedmsg_meta_fedora_infrastructure-openid_libravatar
```

ralphbean:
@Ghost-script, I'm going to go ahead and merge your changes, since they look good and I can't fathom how they'd be directly affecting those bodhi conglomerator tests.  There are other pull requests pending which should hopefully help sort out those issues once everything is all merged in (see #223 and  fedora-infra/fedmsg#325).

PR #222, opened 2015-03-09
Add processor for new karma messages.

ralphbean:
Those get added in fedora-infra/supybot-fedora#22


pypingou:
One file is named ``karma`` the other ``zodbot`` should we name them the same?

ralphbean:
I guess it is a little confusing.

The ``tests/zodbot.py`` file tests both meetbot stuff *and* karma stuff.

But the implementation of those two things is split into two files:  ``meetbot.py`` and ``karma.py``.  They need to be separate processors so that they can have different regexes under the hood.

pypingou:
:+1: for the code now, but maybe it's worth considering splitting the tests into two files as well :)

Issue #226, opened 2015-03-14
Handle multiply-mapped anitya updates


ralphbean:
See this message:

https://apps.fedoraproject.org/datagrepper/id?id=2015-b2e3fab5-12a6-47b0-9ffc-c21d5789a2d0&is_raw=true&size=extra-large

It is a message from anitya where the upstream is mapped to *two* different packages in Fedora.  Our ``subtitle`` text should convey that somehow.

Ghost-script:
@ralphbean , something like this will be okay? 
`A new version of "SQLAlchemy" has been detected: "1.0.0b1" in advance of "0.9.9" , mapped to packages "python-sqlalchemy" and "python-sqlalchemy0.5"`

ralphbean:
Yeah, I like the first of the two options.  :)

Ghost-script:
@ralphbean , okay. what about mapped to single package?
currently if a package is mentioned then instead of project name package name is used. https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure/blob/develop/fedmsg_meta_fedora_infrastructure/anitya.py#L127-L133

so for single package should i change to

`A new version of "SQLAlchemy" has been detected: "1.0.0b1" in advance of "0.9.9" , mapped to package "python-sqlalchemy"` 

or leave it as it is?

ralphbean:
@Ghost-script I think either approach is fine.  If it feels like implementing one way over another is easier.. go for that.

Ghost-script:
This issue has been fixed in #228, so closing it.

PR #228, opened 2015-03-16
Adding Anitya tests for new version of packages detected mapped to multiple packages

Ghost-script:
@ralphbean , as discussed in issue #226 , i made changes for singly and multiple mapped packages. Had to make changes to some older testcases for this. Please review them. Thanks!

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/anitya.py
+++ b/fedmsg_meta_fedora_infrastructure/anitya.py
@@ -22,7 +22,7 @@
 
 from fasshim import gravatar_url, gravatar_url_from_email, email2fas
 from fedmsg_meta_fedora_infrastructure import BaseProcessor
-
+import fedmsg.meta.base
 
 class AnityaProcessor(BaseProcessor):
     topic_prefix_re = 'org\\.release-monitoring\\.(dev|stg|prod)'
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/anitya.py
```diff
@@ -124,24 +124,33 @@ def subtitle(self, msg, **config):
...
                 if package['distro'] == 'Fedora':
-                    project = package['package_name']
-
+                    packages.append(package['package_name'])
+            
+            if len(packages) == 1:
+                packages = packages[0]
+                text = "package"
+            else:
+                packages = " and ".join([", ".join(packages[0:-1]),packages[-1]])
```
ralphbean: Ooo.. there's a nice utility you can use here if you want.  The decision is yours, though:

```python
import fedmsg.meta.base
packages = fedmsg.meta.base.BaseConglomerator.list_to_series(packages)
```

Ghost-script: @ralphbean,  we can use that function,  not a bad idea, but for `text = "package"`/`text="packages"` we will end up using the `if..else..`.  So what i am saying is if we use `list_to_series` we can avoid the `if...else...` but then the `text` will be fixed (`package`) for single or multiple packages. Otherwise if we are going to use `if..else..` for deciding the `text` then its best as it is, calling `list_to_series' will be overhead.

ralphbean:
Agreed.  Maybe the text could be changed to something like:

```
A new usptream version "SQLAlchemy" has been detected: "1.0.0b1" in advance of
"0.9.9", packaged as python-sqlalchemy and python-sqlalchemy0.5
```

Ghost-script:
@ralphbean , AFAIR `N=3` is default. So i will remove N.

ralphbean:
Cool.  :+1: to merge here :)

PR #232, opened 2015-03-19
Add more info about why karma was given

ralphbean:
...if possible.

This info got added later in fedora-infra/supybot-fedora#24


pypingou:
Cool, :+1:

PR #234, opened 2015-03-22
Be careful with the trac ticket summary.

ralphbean:
Sometimes its not there.  For example:

https://apps.fedoraproject.org/datagrepper/id?id=2015-e3faf510-5f39-4e5f-821c-23583d2c9072&is_raw=true&size=extra-large


pypingou:
:+1:

PR #235, opened 2015-03-22
Strip out None values from the bodhi usernames list.

ralphbean:
In some cases, we're getting None values
here.. I think from the 'agent' field of an anonymous bodhi comment.


pypingou:
:+1:

PR #236, opened 2015-03-23
Add long_form for koji scratch builds.

ralphbean:
We already do this for real builds.  Might as well add details for scratch
builds.

Fixes fedora-infra/fmn#74.


pypingou:
:+1: for me

PR #237, opened 2015-03-24
De-duplicate subtitles from long_form representations.

ralphbean:
Before this, we would include the 'subtitle' as a part of every long_form
representation and it just seems kind of silly now.  It means that, for emails
there would be the Subject with the subtitle, then the contents of the email
would contain the subtitle *again* followed by the long_form.  This gets rid of
that duplicate in there.

We'll have to make one other change to fmn.consumer to tell it to **do**
included the subtitle if the messages show up in a batch email, since the
'Subject' won't be there to clarify.


pypingou:
:+1: for me, let's see what travis says :-)

PR #240, opened 2015-03-29
Adjust docstring to represent the action

pypingou:


ralphbean:
Should the subtitle say that it is older explicitly?

pypingou:
The problem that I have is that the new version may be older, or it may be newer and our algorithm to compare versions failed or upstream is managing two branches at the same time. So I went for the "safe" way on the subtitle (less so on the docstring).

ralphbean:
Right, right.  Agreed.  :+1: then.  :)

PR #244, opened 2015-03-30
Be careful with a null host from koji.

ralphbean:
Fixes fedora-infra/fmn#84.


pypingou:
:+1: for me

ralphbean:
- Also fixes fedora-infra/fmn#78.
- Also fixes fedora-infra/fmn#71.

PR #250, opened 2015-04-14
Add tox config for fun and profit.

ralphbean:


pypingou:
:+1: let's see how this goes :)

PR #251, opened 2015-04-14
Handle some new kinds of bodhi messages

### LINKED ISSUE #96: Use msg2long_form api for bodhi backend stuff
We would use it for bodhi when the masher finishes in order to generate a big templated email.

We could use mako templates for it behind the scenes.

ralphbean:
Fixes #96.

/cc @Ghost-Script


lmacken:
Looks good, :+1: 


:family: :movie_camera:

PR #259, opened 2015-04-24
FAF (ABRT server) processor with tests

mbrysa:


ralphbean:
Looks good :)  I have two suggestions:

- Can you make the ``__icon__`` on the processor and the ``expected_icon`` on the test be this:  https://apps.fedoraproject.org/packages/images/icons/abrt.png
- Can you make the ``secondary_icon`` be ``https://apps.fedoraproject.org/packages/images/icons/{package}.png`` where ``{package}`` is the name of the *first* package in the components list?

This will make the messages end up looking nicer in the [datagrepper feed](https://apps.fedoraproject.org/datagrepper/raw).  If you have other better ideas for the icons, by all means use those.

mbrysa:
Good suggestions, updated.

ralphbean:
Good stuff.  :+1: to merge.

PR #261, opened 2015-04-27
Sometimes, there is no blog post title.

ralphbean:


pypingou:
Sad but true although kinda odd as well :)  :+1:

PR #262, opened 2015-05-01
Try to make admin actions more understandable.

ralphbean:
Just using the action id in the subtitle seems to be confusing everyone who
gets notifications about these via FMN (especially when they can't use the
action id in a URL since that view is restricted to ``@is_admin`` in pkgdb).

This hopefully makes them more readable and should fix fedora-infra/fmn#85.


pypingou:
:+1:

PR #269, opened 2015-05-18
Fix pagure regex.

ralphbean:


ralphbean:
Pagure's fedmsg stuff is working now.  It has its own bus, just like anitya.  You can get messages by editing ``/etc/fedmsg.d/endpoints.py`` and adding the pagure public relay:

```python
config = dict(
    # This is a dict of possible addresses from which fedmsg can send
    # messages.  fedmsg.init(...) requires that a 'name' argument be passed
    # to it which corresponds with one of the keys in this dict.
    endpoints={
        # These are here so your local box can listen to the upstream
        # infrastructure's bus.  Cool, right?  :)
        "fedora-infrastructure": [
            #"tcp://hub.fedoraproject.org:9940",
            #"tcp://stg.fedoraproject.org:9940",
        ],
        #"debian-infrastructure": [
        #    "tcp://fedmsg.olasd.eu:9940",
        #],
        #"anitya-public-relay": [
        #    "tcp://release-monitoring.org:9940",
        #],
        "pagure-public-relay": [
            "tcp://stg.pagure.io:9940",
            "tcp://pagure.io:9940",
        ],
    },
)
```

After the freeze, we can make the fedora infrastructure bus subscribe to this so we get all the messages in the rest of our infra places.

pypingou:
Thanks for this :+1:

PR #273, opened 2015-06-02
Fix problems with pagure processor and test suite.

ralphbean:


pypingou:
Thanks! :) :+1:

ralphbean:
np!  now they should show up in the docs too

PR #274, opened 2015-06-06
Try to avoid pagure exceptions for some unhandled message type.

ralphbean:
This is just a temporary workaround until we can figure out exactly what
message type is unhandled here.  There must be some message type from pagure
that is not being tested in the test suite here.  IMHO, we can fix that in a
separate pull request once we know more.


pypingou:
:+1: for the fix, and I wonder if the message are not ``request`` instead of ``pullrequest`` but I didn't actually check the code to confirm

ralphbean:
Looks like it is ``pullrequest`` here.

PR #283, opened 2015-06-23
Link directly to pagure comments.

ralphbean:


pypingou:
We'll need to improve this but as we can't for now, :+1: for me :)

PR #294, opened 2015-07-07
Tagger conglomerator (for fedora-hubs)

ralphbean:


fedmsg_meta_fedora_infrastructure/tests/conglomerate/tagger/test_tags.py
```diff
@@ -0,0 +1,1134 @@
+import fedmsg.tests.test_meta
+
+
+class TestTaggerConglomerateByUser(
+        fedmsg.tests.test_meta.ConglomerateBase):
+    expected = []
+    expected =[
```
pypingou: Declared twice?

fedmsg_meta_fedora_infrastructure/conglomerators/tagger/tags.py
```diff
@@ -0,0 +1,54 @@
+import fedmsg.meta.base
+from fedmsg_meta_fedora_infrastructure.fasshim import avatar_url
+
+class AbstractTaggerConglomerator(fedmsg.meta.base.BaseConglomerator):
+
+    def merge(self, constituents, **config):
+        ms = constituents  # shorthand
+
+        user = constituents[0]['msg']['user']['username']
```
pypingou: Here we use ``constituents``, below we use ``ms``, we're not being very consistent here

pypingou:
Couple of comments but no blockers (although I would like to harmonize the use of ``ms``/``constituents``).

fedmsg_meta_fedora_infrastructure/tests/conglomerate/tagger/test_tags.py
ralphbean: Oop!  Good catch.

PR #295, opened 2015-07-08
Fix incorrect key.

ralphbean:
See [this message](https://apps.stg.fedoraproject.org/datagrepper/id?id=2015-059f6521-b5e9-47cb-a09d-70b0a1319a10&is_raw=true&size=extra-large) for an example.


pypingou:
wfm, :+1:

PR #297, opened 2015-07-09
Protect ourselves from lists.

ralphbean:


pypingou:
:+1: (or do we want to iterate through the component of the list? I guess they are coming from conglomerate no?)

ralphbean:
> (or do we want to iterate through the component of the list? I guess they are coming from conglomerate no?)

No, they are weird koji internal stuff.  In this case, specific to the ``createImage`` task.

ralphbean:
> they are weird koji internal stuff

I'd add that we likely started hitting this due to a new release of koji that was added to the repos(?)

pypingou:
ok thanks for the explanations

PR #299, opened 2015-07-25
Fix #96  "in advance of" should read "newer than"

vhalli:

PR #300, opened 2015-07-27
Be still more careful with this mm2 field.

ralphbean:
This should fix the bug introduced in #295 that @vhalli and @inheritedburrp hit in #299.


pypingou:
:+1: for me

ralphbean:
Thanks Pierre!

PR #304, opened 2015-08-07
Be **extra** careful.

ralphbean:
This is breaking a datagrepper query currently:

https://apps.fedoraproject.org/datagrepper/raw?topic=org.fedoraproject.prod.fedimg.image.test


pypingou:
Nice pun :) :+1:

ralphbean:
Thanks!

PR #318, opened 2015-09-17
Add and test an infragit processor.

ralphbean:
@relrod wrote a proof of concept of a git hook for our infrastructure git
repos.  We tested it with the puppet repo and [it worked](https://apps.fedoraproject.org/datagrepper/id?id=2015-ee126d0a-a89b-4a09-ba15-4ca3382eb8fe&is_raw=true&size=extra-large)

With this processor merged and deployed, we should be able to add the hook for
our others (like dns, infra-docs, ansible, etc...).

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/infragit.py
+++ b/fedmsg_meta_fedora_infrastructure/infragit.py
@@ -1,0 +1,79 @@
+# This file is part of fedmsg.
+# Copyright (C) 2015 Red Hat, Inc.
+#
+# fedmsg is free software; you can redistribute it and/or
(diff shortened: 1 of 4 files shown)


pypingou:
hm, how is this different from the other git hook/processor we already support?

ralphbean:
> how is this different from the other git hook/processor we already support?

It is very very similar, but it has a different link attribute.

Also, by being a new type of message (new topic) the processor here doesn't have to implement all kinds of backwards compatibility with screwy messages from back in 2012.

pypingou:
:+1: then

PR #321, opened 2015-10-01
Handle new "reason" from hotness.

ralphbean:
It got introduced in fedora-infra/the-new-hotness#76.


pypingou:
:+1: nice and easy

PR #322, opened 2015-10-02
Change the wording from packager owner to maintainers

pypingou:
This for two reasons:
- Any maintainer can change the monitoring flag
- We are trying to reduce the concept of ownership for the one of maintainers
and point of contact


ralphbean:
Sure, :+1:

PR #324, opened 2015-10-16
Use aliases for bodhi links

ralphbean:
...in the last few places where its missing.


pypingou:
We're loosing backward compat here no?

Otherwise :+1:

PR #325, opened 2015-10-18
Various fixes for autocloud.

ralphbean:
- Some grammar fixes.
- Removed duplicate definition of the link() method.
- Changed the link() method to return a link to the autocloud results (instead
  of the koji image_url) **if** the job_id is present.  It appears only in
  newer autocloud messages.
- Include the name of the release (rawhide, 23, 22, etc..) in the subject.
  This helps with human parsing of the datagrepper results.


pypingou:
:+1:

PR #334, opened 2015-10-20
If only status is updated, show status in subtitle

puiterwijk:


ralphbean:
LGTM, :+1:

PR #335, opened 2015-10-20
Add atomic icon for atomic messages.

ralphbean:
Eventually, when we get more different kinds of releng messages, we'll want to distinguish between atomic and other stuff.. but this works for now.


puiterwijk:
:+1:

PR #336, opened 2015-10-28
Add support and test for anitya's messages about flagged projects

pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/183

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/tests/anitya.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/anitya.py
@@ -868,7 +868,7 @@ class TestRemoveDistro(Base):
     }
 
 
-class TestProjecFlag(Base):
+class TestProjectFlag(Base):
     """ These messages are published when an user *flags* a project
     in the `anitya <https://release-monitoring.org>`_ database to ask an
     admin to do something on a project.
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/tests/anitya.py
```diff
@@ -867,6 +868,55 @@ class TestRemoveDistro(Base):
...
+              "homepage": "http://kassiopeia.juls.savba.sk/~garabik/software/grc.html",
+              "id": 5777,
+              "backend": "Github"
+            },
+            "message": {
+              "project": "generic-colouriser",
+              "reason": "Delete in favor of the correctly configured 7894.",
+              "agent": "pingou@fedoraproject.org"
+            },
+            "distro": None
```
ralphbean: If ``distro`` is not ``None``, what will it be?  If it is present, can we figure out what Fedora package the project maps to, and return that in ``def packages``?  That would let these messages then be routed through FMN to package maintainers.

pypingou: I've been going through recent messages in datagrepper and all have the distro set to None, makes me wonder if that's even used anywhere

pypingou: Ok so looking at the tests, it's only used when something is actually happening for a distro (adding, editing, deleting)

ralphbean: OK.  Leveraging the mappings isn't strictly necessary here, but it would be nice if we could notify Fedora maintainers when their upstream projects get flagged.

pypingou: Agreed, looking into it, it seems that the fedmsg message ( https://apps.fedoraproject.org/datagrepper/id?id=2015-4f046ac5-c2fa-43f0-a1b1-84e9c0e5c9da&is_raw=true&size=extra-large)  about project that is mapped ( https://release-monitoring.org/project/1970/ ) do not include the packages information, so that would be the first thing to fix

ralphbean:
:+1: here assuming that the test suite is passing.

pypingou:
Tests are passing \ó/

Merging

PR #339, opened 2015-11-02
Test and fix mailman3 archived-at links that are surrounded with arrow braces.

ralphbean:


pypingou:
:+1: sorry for the lengthy discussion, I had missed the check if archived_at starts with http :)

ralphbean:
All good.  Teamwork makes the dream work!

PR #344, opened 2015-12-08
Add some preliminary processors for the new pungi-koji compose stuff.

ralphbean:


lmacken:
LGTM, aside from the Travis failures.

pypingou:
:+1: for me as well

ralphbean:
A note for the future... that ``/rawhide/`` string hardcoded in the ``link()`` path won't do.  It will break when we moved to "branched" mode in a couple months.

However, there's no way to guess what that string will be right now afaik.  We're going to need to patch pungi-koji to publish that string inside the original message payload.

ralphbean:
OK, that new field should be coming in https://pagure.io/pungi/pull-request/93 and https://pagure.io/pungi/pull-request/95

ralphbean:
And... here's an example message from that:  https://apps.fedoraproject.org/datagrepper/id?id=2015-feaf5f4a-a390-4835-bf0a-4767686b74c4&is_raw=true&size=extra-large

PR #347, opened 2015-12-14
Remove the "markup" stuff.

ralphbean:
It never really got implemented and I think I have a better, more scalable way
to do it in fedora-hubs.


pypingou:
:+1: for me

PR #349, opened 2015-12-16
Return package-centric information about taskotron events.

ralphbean:
We should definitely get some of the QA people to comment in here before merging.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/taskotron.py
+++ b/fedmsg_meta_fedora_infrastructure/taskotron.py
@@ -16,8 +16,15 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 #
 # Authors:  Martin Krizek <mkrizek@redhat.com>
-#
+#           Ralph Bean <rbean@redhat.com>
+
+import logging
+
+import requests
+
 from fedmsg_meta_fedora_infrastructure import BaseProcessor
+
+log = logging.getLogger('fedmsg.meta')
 
 
 class TaskotronProcessor(BaseProcessor):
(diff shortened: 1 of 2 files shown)


ralphbean:
A note about impact here, already discussed in IRC.

- We have a rule in FMN that forwards events to packagers about "packages that they own".
- The taskotron messages to not produce any metadata about what package they are about.
- Therefore FMN does not forward them to anyone.

This changeset gives the fedmsg.meta module knowledge about how to extract the set of packages relevant to different taskotron messages.. which should get *basic* FMN forwarding working.

pypingou:
:+1: for me

mkrizek:
:+1:  LGTM

fedmsg_meta_fedora_infrastructure/taskotron.py
```diff
@@ -39,5 +46,45 @@ def link(self, msg, **config):
...
+                package,
+                msg['msg']['result'].get('outcome', '')
+            ])
+            for package in packages
+        ])
+    def packages(self, msg, **config):
+        type = msg['msg']['task']['type']
+        if type == 'koji_build':
+            nvr = msg['msg']['task']['item']
+            name, version, release = nvr.rsplit('-', 2)
```
kparal: I'm a bit uneasy when doing NEVR operations by hand, that's why we use low level tools for it (`rpmUtils` from yum in the old days, `hawkey.split_nevra()` nowadays). It would break if we sent you ENVR string (yum style), but I think we don't do it, we tried to convert everything to NEVR (dnf style) or NVR (because some services don't process epoch correctly).

fedmsg_meta_fedora_infrastructure/taskotron.py
```diff
@@ -39,5 +46,45 @@ def link(self, msg, **config):
...
+            if not bool(resp):
+                log.warn("Failed to talk to bodhi %r %r" % (resp, resp.url))
+                return set()
+            data = resp.json()
+            builds = data['update']['builds']
+            nvrs = [build['nvr'] for build in builds]
+            packages = [nvr.rsplit('-', 2)[0] for nvr in nvrs]
+            return set(packages)
+
+        log.warn('Unhandled taskotron type %r' % type)
```
kparal: If you see this, please let us know and we'll send a patch. But currently I think we only publish `koji_build` and `bodhi_update`.

kparal:
Thanks a lot for the patch, @ralphbean. I'm not able to verify whether it works, but it looks good to me.

fedmsg_meta_fedora_infrastructure/taskotron.py
ralphbean: Acknowledged.  If resultsdb eventually starts including a list of packages along with its fedmsg payload, this code can be revised to drop the NVR parsing and use that list instead.

fedmsg_meta_fedora_infrastructure/taskotron.py
ralphbean: Will do.  Almost all of the processors in this module have some kind of warning at the end like this.

ralphbean:
> I'm not able to verify whether it works, but it looks good to me.

The test suite works.  :)  Which is a good, if not sure-fire sign.

PR #355, opened 2016-02-01
Handle a whole slough of new squirely releng messages.

ralphbean:
They all come from this script:

https://pagure.io/releng/blob/master/f/scripts/make-updates

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/compose.py
+++ b/fedmsg_meta_fedora_infrastructure/compose.py
@@ -59,12 +59,54 @@ class ComposeProcessor(BaseProcessor):
         elif msg['topic'].endswith('.pungify.complete'):
             tmpl = self._(
                 "finished building boot.iso for {branch}{arch}")
+
+        elif msg['topic'].endswith('make-updates.start'):
+            tmpl = self._("started a run of F{branch} make-updates")
+        elif msg['topic'].endswith('make-updates.start'):
+            tmpl = self._("finished a run of F{branch} make-updates")
+
+        elif msg['topic'].endswith('cloudimg-build.start'):
+            tmpl = self._("started the cloudimg-build phase of "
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/compose.py
```diff
@@ -76,9 +118,19 @@ def link(self, msg, **config):
         else:
             if 'epelbeta' in msg['topic']:
                 base = "https://dl.fedoraproject.org/pub/epel/beta/7"
-            else:
+            elif 'branched' in msg['topic'] or 'rawhide' in msg['topic']:
                 base = "https://dl.fedoraproject.org/pub/" + \
                     "fedora/linux/development"
+            else:
+                branch = msg['msg'].get('branch', 'not-a-number')
```
pypingou: is this rather 'not-a-branch'?

pypingou:
:+1: but looks a little weak on tests, including if the branch isn't specified or is not a number

fedmsg_meta_fedora_infrastructure/compose.py
ralphbean: No - "not-a-number", because I try to cast it to an ``int`` just two lines below.

ralphbean: The not-a-number could be "rawhide", for instance.  ;)

ralphbean:
As far as I can tell, that make-updates script always provides the branch and it is always a number.  If it is absent or not a number, I think it will fail back to the other code paths just fine (and the other tests test those code paths).  Thanks @pypingou!

fedmsg_meta_fedora_infrastructure/compose.py
pypingou: But then ``msg['msg'].get('branch')`` would return `rawhide` no?

PR #358, opened 2016-02-29
Handle new mdapi format.

ralphbean:
As discussed in IRC.


fedmsg_meta_fedora_infrastructure/tests/mdapi.py
```diff
@@ -26,10 +27,8 @@
 from .common import add_doc
 
 
-class TestNewProject(Base):
-    """ These messages are published when a new project is created on
-    `pagure <https://pagure.io>`_.
-    """
+class TestNewProjectLegacy(Base):
+    """ These are whta the messages used to look like... """
```
pypingou: typo

pypingou:
Looks good :+1: Let's hope we don't have a revision 3, otherwise we'll end-up with new_new_school, old_new_school and old_old_school :D

PR #359, opened 2016-03-01
Fix situation where pagure messages have zero comments.

ralphbean:
Sometimes, apparently, the comments entry here can be empty, so we would get
"index out of range".


pypingou:
:+1: for me

Looks like the issue might be triggered by new issues or PR being created: https://apps.fedoraproject.org/datagrepper/id?id=2016-fcc6f2bd-4f5c-4029-9795-9c7af8325c7a&is_raw=true&size=extra-large

PR #364, opened 2016-03-09
Fix copr links.

### LINKED ISSUE #363: Broken Copr log links
In the emails about copr builds, the URL for build log is: https://copr-be.cloud.fedoraproject.org/results/msimacek/mock/fedora-24-x86_64/mock/build.log.gz
Accessing that, I get 404.
The URL I get when looking up the logs from copr's web UI is:
https://copr-be.cloud.fedoraproject.org/results/msimacek/mock/fedora-24-x86_64/00166792-mock/build.log
(which works)

I think the processor should be updated to point to current URLs.

ralphbean:
Apparently, they changed directory structure on copr-be.

This should address that and fix #363.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/coprs.py
+++ b/fedmsg_meta_fedora_infrastructure/coprs.py
@@ -40,11 +40,11 @@ Status:   {status}
 ID:       {build}
 
 Logs:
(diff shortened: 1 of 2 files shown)


pypingou:
:+1: for me

ralphbean:
Thanks @pypingou!

PR #366, opened 2016-03-16
Pad against a race condition with koji.

ralphbean:
We *often* hit a race condition with koji here.

It publishes the fedmsg message before it has committed the details to its
database.  When we receive the message, we here try to query koji for more
details but... the fedmsg got to us before the db commit finished and koji says
"I know nothing."


pypingou:
:+1:

PR #368, opened 2016-03-17
Long form for planet.

ralphbean:
You can see how this plays out in the screenshots over at https://pagure.io/fedora-hubs/pull-request/122


pypingou:
Cool :+1: and looks nice!

PR #369, opened 2016-03-22
Conglomerate pagure events.

ralphbean:
Around tickets, PRs, and users who commit.


fedmsg_meta_fedora_infrastructure/conglomerators/pagure/pagure.py
```diff
@@ -0,0 +1,139 @@
+import fedmsg.meta.base
+from fedmsg_meta_fedora_infrastructure.fasshim import avatar_url, email2fas
+
+
+
+
```
pypingou: Still want to follow pep8? :)

fedmsg_meta_fedora_infrastructure/conglomerators/pagure/pagure.py
```diff
@@ -0,0 +1,139 @@
...
+        tmpl['link'] = link_template.format(
+            repo=repo,
+            entity_name=self.entity_name,
+            idx=idx,
+        )
+
+        return tmpl
+
+
+class ByPR(AbstractPagureTicketConglomerator):
```
pypingou: I like this approach with the parent class inherited by each conglomerators :)

pypingou:
:+1: for me but do I read this correctly that we don't have tests for it? Is that intended?

ralphbean:
> do I read this correctly that we don't have tests for it?

Nope.  They are here:  https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure/blob/feature/pagure-conglomeration/fedmsg_meta_fedora_infrastructure/tests/conglomerate/pagure/test_pagure.py

github says that "the diff is to large to show here".

fedmsg_meta_fedora_infrastructure/conglomerators/pagure/pagure.py
ralphbean: Yah, me too!

ralphbean:
Thanks @pypingou!

PR #370, opened 2016-03-24
add openQA processor

AdamWill:
I don't think the tests are running correctly locally (every 'expected' result comes back as an empty string) so @ralphbean said to just send a PR and we'll work it out here. Currently openQA fedmsg's are quite basic so there's only so much processing we can do. Future possibilities include indicating more properties of the job (test suite name, arch, image under test?), result for finished jobs (this is an upstream issue), stuff like that. We don't have a 'who' for openQA events, really, because they're most commonly triggered automatically.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/openqa.py
+++ b/fedmsg_meta_fedora_infrastructure/openqa.py
@@ -1,0 +1,64 @@
+# This file is part of fedmsg.
+# Copyright (C) 2016 Red Hat, Inc.
+#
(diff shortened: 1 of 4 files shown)


AdamWill:
OK, so now I fixed all the dumbass syntax errors the tests actually pass! Funny how that works...

ralphbean:
:+1: here, this looks good.  Thanks for taking the time to do this!

AdamWill:
thanks ralph!

PR #372, opened 2016-03-30
Handle new compat issues with copr messages.

ralphbean:
The message format changed (the `pkg` key is now just the package name, not the
whole NVR).

This updates one of the tests to ensure that we handle both the new and the old
format.


pypingou:
Cool, :+1:

PR #373, opened 2016-03-31
Add collection field to koschei processor

msimacek:
Adding a human readable collection name to koschei message. The change should be compatible both ways.


ralphbean:
Looks good to me.  :+1: to merge.

PR #374, opened 2016-03-31
Add a PDC processor with tests.

ralphbean:


msimacek:
The processor looks ok to me. But PDC seems to be very slow in responding to the API links (like https://pdc.fedoraproject.org/rest_api/v1/composes/)

ralphbean:
Yeah, It's trying to return too much information.  I'll bury those links... :/

PR #375, opened 2016-04-04
Return the package associated with a bz update.

ralphbean:


fedmsg_meta_fedora_infrastructure/bz.py
```diff
@@ -145,6 +150,12 @@ def _gather_emails(self, msg):
 
         return users
 
+    def packages(self, msg, **config):
+        component = msg['msg']['bug']['component']
+        if component in bz_components_that_are_not_packages:
+            return set()
```
pypingou: Shouldn't we try to extract the package name then? Maybe from the title?

ralphbean: hotness has had a lot of trouble trying to do that :(

We could try.. but lots of people use different/non-standard formats when they type in the info.

pypingou: ok let's forget it then

pypingou:
:+1: for me

ralphbean:
Thanks @pypingou!

PR #376, opened 2016-04-06
openQA: revise meta processor for pending new message format

AdamWill:
I'm currently working on changing the openQA fedmsg message
format to add new information and improve consistency a bit:
https://github.com/os-autoinst/openQA/pull/617
Those changes will hit Fedora soon, but first we have to make
things ready for them. This adjusts the meta processor such
that it should produce more informational messages for the new
format while still working with the old format, and extends
the tests to cover both cases.

Note: I'm not sure exactly how the magic that turns test docstrings into message topic docs works, so I don't know if `TestOpenQAHdd1`'s docstring will be a problem there. If so, please let me know how to fix it :) Thanks!

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/doc_utilities.py
+++ b/fedmsg_meta_fedora_infrastructure/doc_utilities.py
@@ -157,6 +157,10 @@ def make_topics_doc(output_dir):
             if 'Legacy' in cls.context.__name__:
                 continue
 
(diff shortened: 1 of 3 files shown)


ralphbean:
I think this works as-is.  It's fine to have that appear in the docs.  ;)

ralphbean:
:+1: thanks!

PR #378, opened 2016-04-07
Handle some namespaced scm things.

ralphbean:


fedmsg_meta_fedora_infrastructure/scm.py
```diff
@@ -193,6 +202,12 @@ def usernames(self, msg, **config):
             return set()
 
     def packages(self, msg, **config):
+
+        # If its not an rpms/ thing, then its not a package.
+        if 'namespace' in msg['msg']:
+            if msg['msg']['namespace'] != 'rpms':
+                return set()
+
```
pypingou: Should we have some specific notifications/conversion for the ``*-checks`` repo? I figure some people would like to be notified of changes done in them.

pypingou:
One question, but code wise :+1:

fedmsg_meta_fedora_infrastructure/scm.py
ralphbean: Hm, this is a good point.  Will do!

pypingou:
:+1: 

We may want to open a ticket on FMN to add a filter for namespaces. Allowing notification from rpms only, or rpms and rpms-check or docker and rpms & so on.

PR #383, opened 2016-05-11
Perform 26 smaller queries instead of 1 to retrieve the entire list of users

pypingou:
This actually slows down thing a little but at least it no longer runs
into time-out issues.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/fasshim.py
+++ b/fedmsg_meta_fedora_infrastructure/fasshim.py
@@ -1,6 +1,7 @@
 import collections
 import threading
 import socket
+import string
 from hashlib import sha256, md5
 
 _fas_cache = {}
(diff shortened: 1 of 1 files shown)


abompard:
I suppose there is no way to page requests on the server?

pypingou:
Unfortunately not on FAS2

pypingou:
grmbl, wrong button :)

abompard:
OK, :+1:

lmacken:
:+1: Fine with me.

ralphbean:
I added two small mostly cosmetic changes to this PR.

Let's merge it and get it out the door.

sachinkamath:
@ralphbean  +1

Looks good to me :+1:

ralphbean:
:)  Thanks all!

PR #386, opened 2016-06-28
Update Autocloud Processor with the new autocloud message format

sayanchowdhury:


fedmsg_meta_fedora_infrastructure/autocloud.py
```diff
@@ -29,39 +29,66 @@ class AutoCloudProcessor(BaseProcessor):
     __obj__ = "Cloud Image Test"
 
     def subtitle(self, msg, **config):
-        image_name = msg['msg']['image_name']
         status = msg['msg']['status']
-        release = msg['msg'].get('release')
         if 'autocloud.image' in msg['topic']:
+            compose_id = msg['msg']['compose_id']
```
pypingou: Old messages won't have the compose_id and will thus fail

pypingou:
Let's keep/add a test w/ the old format, fedmsg_meta must be compatible with all messages in datagrepper

fedmsg_meta_fedora_infrastructure/autocloud.py
```diff
@@ -21,14 +21,98 @@
...
     __docs__ = "https://github.com/kushaldas/autocloud"
     __icon__ = "https://apps.fedoraproject.org/img/icons/fedimg.png"
     __obj__ = "Cloud Image Test"
 
+
+    def _func_router(self, msg, fname, **config):
+        FNAME = 'handle_%s_autocloud_%s'
+
+        if 'compose_id' in msg['msg'] or 'compose_job_id' in msg['msg']:
+            return getattr(self, FNAME % ('new', fname))(msg, **config)
```
pypingou: Instead of using new/old you may want to do something like v1/v2, otherwise you may end having problems if you change the format once more.

Also, while I find the approach quite neat, do we loose a bit in readability?

sayanchowdhury: Updated the code to v1/v2.

I don't think much of readability is lost as it's just a method call. Also, I added a few lines to explain the methods functionality

pypingou:
LGTM :)

sayanchowdhury:
Thanks :)

PR #388, opened 2016-07-07
Fix mailman3 icon broken link

Prashant-Surya:
Replaced broken link of mailman3 icon


pypingou:
Same here, let's merge :)

PR #389, opened 2016-07-08
replace yum by dnf in readme

vivekanand1101:


skrzepto:
I believe these services run on rhel/centos servers which are still using yum. So not sure if updating readme to dnf is useful here

vivekanand1101:
@skrzepto  you can close the pr if its worthless :)

puiterwijk:
Correct, RHEL still uses yum, and yum will still work for dnf people.

PR #394, opened 2016-07-13
Add the test for the broken messages and the fix for the Autocloud Processor

sayanchowdhury:
Add the test for the broken messages and the fix for the Autocloud Processor

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/autocloud.py
+++ b/fedmsg_meta_fedora_infrastructure/autocloud.py
@@ -112,6 +112,7 @@ class AutoCloudProcessor(BaseProcessor):
             if compose_job_id is not None:
                 return 'https://apps.fedoraproject.org/autocloud/jobs/%s' % (
                     compose_job_id)
+        return 'https://apps.fedoraproject.org/autocloud/'
 
     def handle_v2_autocloud_objects(self, msg, **config):
         status = msg['msg']['status']
--- a/fedmsg_meta_fedora_infrastructure/autocloud.py
+++ b/fedmsg_meta_fedora_infrastructure/autocloud.py
@@ -86,10 +86,8 @@ class AutoCloudProcessor(BaseProcessor):
                                 compose_id=compose_id)
 
         if "autocloud.compose" in msg['topic']:
-            compose_id = msg['msg'].get('id', '')
-
-            if compose_id == '' and "complete" in msg['topic']:
-                compose_id = msg['msg'].get('compose_id', '')
+            compose_id = msg['msg'].get(
+                    'id', msg['msg'].get('compose_id', 'A compose'))
 
             if status == 'queued':
                 tmpl = self._("{compose_id} is {status} for testing")
--- a/fedmsg_meta_fedora_infrastructure/autocloud.py
+++ b/fedmsg_meta_fedora_infrastructure/autocloud.py
@@ -38,7 +38,8 @@ class AutoCloudProcessor(BaseProcessor):
         """
         FNAME = 'handle_%s_autocloud_%s'
 
-        if 'compose_id' in msg['msg'] or 'compose_job_id' in msg['msg']:
+        if ('compose_id' in msg['msg'] or 'compose_job_id' in msg['msg'] or
+                'autocloud.compose' in msg['topic']):
             return getattr(self, FNAME % ('v2', fname))(msg, **config)
         else:
             return getattr(self, FNAME % ('v1', fname))(msg, **config)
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/autocloud.py
```diff
@@ -85,7 +86,10 @@ def handle_v2_autocloud_subtitle(self, msg, **config):
                                 compose_id=compose_id)
 
         if "autocloud.compose" in msg['topic']:
-            compose_id = msg['msg']['id']
+            compose_id = msg['msg'].get('id', '')
+
+            if compose_id is '' and "complete" in msg['topic']:
```
pypingou: Don't you want to use `` == ''`` here?

fedmsg_meta_fedora_infrastructure/autocloud.py
```diff
@@ -85,7 +86,10 @@ def handle_v2_autocloud_subtitle(self, msg, **config):
...
+                compose_id = msg['msg'].get('compose_id', '')
```
pypingou: So what's happening here is compose_id is still ``''`` ?

fedmsg_meta_fedora_infrastructure/autocloud.py
```diff
@@ -106,9 +110,10 @@ def handle_v2_autocloud_link(self, msg, **config):
             else:
                 return msg['msg']['image_url']
         elif 'autocloud.compose' in msg['topic']:
-            compose_job_id = msg['msg']['compose_job_id']
-            return 'https://apps.fedoraproject.org/autocloud/jobs/%s' % (
-                compose_job_id)
+            compose_job_id = msg['msg'].get('compose_job_id')
+            if compose_job_id is not None:
+                return 'https://apps.fedoraproject.org/autocloud/jobs/%s' % (
+                    compose_job_id)
```
pypingou: Is there a default value to return?
If compose_job_id is None what should be returned?

sayanchowdhury: Since I am not returning anything it should return None. If there is no compose_job_id then it should not create any URL.

fedmsg_meta_fedora_infrastructure/autocloud.py
sayanchowdhury: Changed this

pypingou: You changed the ``==`` but did not answer this question :)

sayanchowdhury: If the key `id` is not found then it tries to fetch it from `compose_id`. Fix for this https://github.com/kushaldas/autocloud/pull/44

pypingou: And if none of them are present?

With this code we will not have an error but the message returned will also not be understandable. So should we crash so that we are at least notified of an odd message?

The part that worries me a little is:
* Retrieve ``id``
* If there is no ``id`` *and the topic contains ``complete``*: retrieve compose_id

So if there is no ``id`` but the topic does not contain ``complete`` then ``compose_id`` remains ``''``

sayanchowdhury: The chances are very less (can say nil) that both are missing. Maybe we can add a default text for compose_id as `A compose`.

This code is under another check above if `autocloud.compose` is in the topic. So the resultant check would be like `autocloud.compose` and `complete` in the topic. 
I can change `complete` to `autocloud.compose.complete`

pypingou: +1 for the default

But do we need that check for 'complete', we could just do something like: 
``compose_id = msg['msg'].get('id', compose_id = msg['msg'].get('compose_id', 'A compose'))``
thoughts?

sayanchowdhury: Never thought of that. I have updated the code with the changes.

fedmsg_meta_fedora_infrastructure/autocloud.py
pypingou: We could return simply ``https://apps.fedoraproject.org/autocloud/`` no?

sayanchowdhury: Updated to return `https://apps.fedoraproject.org/autocloud/` as default

pypingou:
:+1: for me if travis agrees :)

sayanchowdhury:
Thanks :)

PR #395, opened 2016-07-26
Query only for active FAS users.

ralphbean:
This is a follow-on to #383 that takes advantage of fedora-infra/fas#162.

It seems that the code from #383 fails now given that spammers have bloated our
FAS db.  Searching just for `'a*'` times out/fails.

By introducing `'status': 'active',` here, it gets things working again.
Querying for just `'a*'` took about 3.5 minutes on a local test at the time of
writing.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/fasshim.py
+++ b/fedmsg_meta_fedora_infrastructure/fasshim.py
@@ -125,7 +125,10 @@ def make_fas_cache(**config):
             log.info("Downloading FAS cache for %s*" % key)
             response = fasclient.send_request(
                 '/user/list',
-                req_params={'search': '%s*' % key},
+                req_params={
+                    'search': '%s*' % key,
+                    'status': 'active',
+                },
                 auth=True)
         except fedora.client.ServerError as e:
             log.warning("Failed to download fas cache for %s %r" % (key, e))


pypingou:
I've been using this locally to make things work, so def :+1: for me! :)

pypingou:
Although one question: what about processing old messages involving accounts that got deactivated?

ralphbean:
> Although one question: what about processing old messages involving accounts that got deactivated?

True, those will fail.  But what specifically will fail?

- For badges.. we don't care about awarding badges to deactivated accounts, right?
- For FMN.. deactivated accounts shouldn't care about receiving notifications, right?
- For hubs.. same thing?

Are there any cases where this consideration has a negative impact on UX?

pypingou:
For our apps it is likely fine.

I do wonder about two options:
* give the user of fedmsg_meta a way to turn off this filter
* figure out a way to retrieve info about a specific user, if we cannot find the info in the cache, so if user ``foo`` is deactivated, it's not in the cache, you're processing a message where it is involved, fedmsg_meta goes query FAS for just that user. For most accounts/messages we would have the same performances, for older ones it would just be a little slower.

What do you think?

sachinkamath:
This is a possible solution for #399 , right?

pypingou:
As far as I can see this shouldn't impact #399 no :(

ralphbean:
Hm.. this has been sitting around for almost a year.  What should we do about it?

pypingou:
Tbh, I don't know, maybe the folks working on hubs would have an opinion?
In a way I'm in favor of this change, on the other side I'm a little concerned about its potential impact.

For one, I think FMN might be impacted/need to be adjusted.

PR #396, opened 2016-07-28
Fix something we faced in FMN stg, copr messages where the agent is None

pypingou:
Having None in the list of agent made the call to self.list_to_series
fail.
With this, we should avoid the issue


ralphbean:
Nice!  :)

PR #397, opened 2016-08-09
Adjust after diff access levels in pagure

vivekanand1101:
There are a couple of changes in the way pagure will send info via fedmsg after [this](https://pagure.io/pagure/pull-request/1177). For groups, i will add them in a diff PR. Also, i have just ran the tests, i don't know if there is any other way to test this locally.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/pagure.py
+++ b/fedmsg_meta_fedora_infrastructure/pagure.py
@@ -284,10 +284,27 @@ class PagureProcessor(BaseProcessor):
             return tmpl.format(user=user, project=project, fields=fields)
         elif 'pagure.project.user.added' in msg['topic']:
             new_user = msg['msg']['new_user']
+            access = msg['msg'].get('access')
+            if access:
+                tmpl = self._(
+                    '{user} added "{new_user}" to project {project} '
+                    'with {access} access'
+                )
+            else:
+                tmpl = self._(
+                    '{user} added "{new_user}" to project {project}'
+                )
+            return tmpl.format(
+                user=user, project=project, new_user=new_user, access=access)
+        elif 'pagure.project.user.access.updated' in msg['topic']:
+            new_user = msg['msg']['new_user']
+            new_access = msg['msg']['new_access']
             tmpl = self._(
-                '{user} added "{new_user}" to project {project}'
+                '{user} updated access of "{new_user}" to {new_access} '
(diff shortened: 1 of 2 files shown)


vivekanand1101:
Just in case someone misses, the access level PR is currently merged in a diff branch than master.

fedmsg_meta_fedora_infrastructure/pagure.py
```diff
@@ -284,10 +284,22 @@ def subtitle(self, msg, **config):
             return tmpl.format(user=user, project=project, fields=fields)
         elif 'pagure.project.user.added' in msg['topic']:
             new_user = msg['msg']['new_user']
+            access = msg['msg']['access']
```
pypingou: This is not backward compatible, while fedmsg_meta needs to be able to handle both new and old format of messages

fedmsg_meta_fedora_infrastructure/tests/pagure.py
```diff
@@ -782,7 +782,7 @@ class TestProjectUserAdded(Base):
     project on `pagure <https://pagure.io>`_.
     """
     expected_title = "pagure.project.user.added"
-    expected_subti = 'pingou added "ralph" to project foo'
+    expected_subti = 'pingou added "ralph" to project foo with admin access'
```
pypingou: We need to keep the current test as is, just modify its class to ``LegacyTestProjectUserAdded`` and create a new ``TestProjectUserAdded`` with the new content

fedmsg_meta_fedora_infrastructure/pagure.py
```diff
@@ -284,10 +284,27 @@ def subtitle(self, msg, **config):
             return tmpl.format(user=user, project=project, fields=fields)
         elif 'pagure.project.user.added' in msg['topic']:
             new_user = msg['msg']['new_user']
+            access = msg['msg'].get('access', None)
```
sayanchowdhury: `get` return `None` by default

vivekanand1101: I am aware of that :p . I have seen it being used a lot in pagure, so i thought it might be helpful for readability. I will change it if you want.

sayanchowdhury:
:+1:  for me

pypingou:
Looks good to me as well, let's merge :)

Thanks!

PR #398, opened 2016-08-10
Add support for project.group.added and project.group.access.updated …

vivekanand1101:
…for pagure messages

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/pagure.py
+++ b/fedmsg_meta_fedora_infrastructure/pagure.py
@@ -282,6 +282,29 @@ class PagureProcessor(BaseProcessor):
             tmpl = self._(
                 '{user} edited the {fields} fields of project {project}')
             return tmpl.format(user=user, project=project, fields=fields)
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/pagure.py
```diff
@@ -282,6 +282,24 @@ def subtitle(self, msg, **config):
             tmpl = self._(
                 '{user} edited the {fields} fields of project {project}')
             return tmpl.format(user=user, project=project, fields=fields)
+        elif 'pagure.project.group.added' in msg['topic']:
+            new_group = msg['msg']['new_group']
+            access = msg['msg']['access']
```
pypingou: Hm, we do have messages already in datagrepper that do not have ``access``: https://apps.fedoraproject.org/datagrepper/id?id=2016-6be0ffdb-dbe0-452b-bd1c-4868460675d2&is_raw=true&size=extra-large

So we need to support these as well

pypingou:
Perfect! thanks! :)

PR #400, opened 2016-09-09
When copr repo belongs to a group, alter the link

hroncok:
Fixes https://github.com/fedora-infra/fmn/issues/134

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/tests/coprs.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/coprs.py
@@ -253,6 +253,10 @@ class TestCoprsGroupLink(Base):
     """ `Copr <https://fedorahosted.org/copr/>`_ publishes these messages
     when a build has completed.  This is an example of a group repo.
     """
+    expected_title = "copr.build.end"
+    expected_subti = (
+        "churchyard's python26 copr build of python26-2.6.9-1.fc26 for "
+        "fedora-23-i386 finished with 'failed'")
     expected_link = ("https://copr.fedoraproject.org/coprs/"
                      "g/python/python26/build/450811/")
 
--- a/fedmsg_meta_fedora_infrastructure/tests/coprs.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/coprs.py
@@ -249,6 +249,39 @@ class TestCoprsWorkerCreate(Base):
     }
 
 
+class TestCoprsGroupLink(Base):
+    """ `Copr <https://fedorahosted.org/copr/>`_ publishes these messages
+    when a build has completed.  This is an example of a group repo.
+    """
+    expected_link = ("https://copr.fedoraproject.org/coprs/"
+                     "g/python/python26/build/450811/")
+
+    {
+        u'source_name': u'datanommer',
+        u'i': 3,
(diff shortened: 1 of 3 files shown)


hroncok:
I just edited this on GitHub, but it seems test are failing, so I'll look into that later

pypingou:
This is the error:
````
  File "/home/travis/build/fedora-infra/fedmsg_meta_fedora_infrastructure/fedmsg_meta_fedora_infrastructure/coprs.py", line 132, in link
    if user.startswith('@'):
AttributeError: 'NoneType' object has no attribute 'startswith'
````
Should be easy to fix :)

**Comment by hroncok** [comment]
Thanks, I've seen it. I could only do it if it is a string... Not sure if that's the solution and why it is None in the first place.

**Comment by pypingou** [comment]
the simplest would be: ```  if user and user.startswith('@'):```

**Comment by hroncok** [comment]
You were seconds faster :)

**Comment by pypingou** [comment]
Looks good to me, let's merge! :)

**Comment by pypingou** [comment]
hm actually, it looks good but I think we should add unit-tests for it, to be on the safe side of things

**Comment by hroncok** [comment]
In the spirit of the original Pull Request, I'll try to provide test only by editing trough GitHub web interface.  Any struggles can later be squashed :)

**Comment by hroncok** [comment]
Seems to work.

File: fedmsg_meta_fedora_infrastructure/tests/coprs.py [hunk 1]
```diff
@@ -249,6 +249,39 @@ class TestCoprsWorkerCreate(Base):
     }
 
 
+class TestCoprsGroupLink(Base):
+    """ `Copr <https://fedorahosted.org/copr/>`_ publishes these messages
+    when a build has completed.  This is an example of a group repo.
+    """
+    expected_link = ("https://copr.fedoraproject.org/coprs/"
```
pypingou: Could we include the expected title as well?

hroncok: I don't know what that's expected to be. `copr.build.end`?

pypingou: Indeed, the title isn't really informative, the subtitle is though: 
https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure/blob/develop/fedmsg_meta_fedora_infrastructure/tests/coprs.py#L103-L104

pypingou: So should we add the subtitle?

hroncok:
Travis build is broken not by my change I believe.

fedmsg_meta_fedora_infrastructure/tests/coprs.py
hroncok: done

pypingou:
> error: Setup script exited with error in urllib3 setup command: Invalid environment marker: python_version <= "2.7"

...

> [31;1mThe command "python setup.py install" failed and exited with 1 during .[0m

Bummer :(

hroncok:
Last commit fixes the issue.

pypingou:
Cool! Looks good to me, let's merge :)

PR #404, opened 2016-10-24
Use http for koji, it does not use a well-known CA

tyll:


ralphbean:
I think @puiterwijk may be working on some cert things for koji.fp.o, which might be relevant to a discussion here.

puiterwijk:
This has been in for so long now that I'm not feeling we should change it now, given that we're just about to fix that soon. (I've heard Dec 12 is the day we switch koji to a real CA cert).

pypingou:
I'm going to close this as I believe it no longer applies.

Thanks for your work though :)

PR #413, opened 2017-02-13
Handle broken Pagure assigned.added message

puiterwijk:


puiterwijk:
The test failures are not caused by my patch as far as I'm able to see.

puiterwijk:
For the record: this change is currently live at production FMN because it halted all message sending.

jeremycline:
@puiterwijk is there a reason this hasn't been merged yet? @ralphbean needs a release out with #417 so we can use https://github.com/fedora-infra/fmn/pull/174, but we can't until this is merged since it's hotfixed.

Issue #123, opened 2014-08-08
traceback with github messages


tyll:
There were tracebacks when using fedmsg_meta_fedora_infrastructure which seem
to be related to github messages:

```
2014-08-07 22:10:25,053: ERROR: Exception: Traceback (most recent call last):
  File "./autosigner.py", line 433, in <module>   
    subtitle = fedmsg.meta.msg2subtitle(msg, **config)
  File "/usr/lib/python2.7/site-packages/fedmsg/meta/__init__.py", line 132, in __wrapper
    return f(msg, **config)
  File "/usr/lib/python2.7/site-packages/fedmsg/meta/__init__.py", line 151, in __wrapper
    return f(msg, processor=processor, **config)
  File "/usr/lib/python2.7/site-packages/fedmsg/meta/__init__.py", line 185, in msg2subtitle
    return processor.subtitle(msg, **config)
  File "/usr/lib/python2.7/site-packages/fedmsg_meta_fedora_infrastructure/github.py", line 76, in subtitle
    user = self._get_user(msg)
  File "/usr/lib/python2.7/site-packages/fedmsg_meta_fedora_infrastructure/github.py", line 34, in _get_user
    user = msg['msg']['commit'].get('author', {}).get('login', {})
AttributeError: 'NoneType' object has no attribute 'get'
2014-08-07 22:10:36,098: ERROR: Exception: Traceback (most recent call last):
  File "./autosigner.py", line 433, in <module>
    subtitle = fedmsg.meta.msg2subtitle(msg, **config)
  File "/usr/lib/python2.7/site-packages/fedmsg/meta/__init__.py", line 132, in __wrapper
    return f(msg, **config)
  File "/usr/lib/python2.7/site-packages/fedmsg/meta/__init__.py", line 151, in __wrapper
    return f(msg, processor=processor, **config)
  File "/usr/lib/python2.7/site-packages/fedmsg/meta/__init__.py", line 185, in msg2subtitle
    return processor.subtitle(msg, **config)
  File "/usr/lib/python2.7/site-packages/fedmsg_meta_fedora_infrastructure/github.py", line 76, in subtitle
    user = self._get_user(msg)
  File "/usr/lib/python2.7/site-packages/fedmsg_meta_fedora_infrastructure/github.py", line 34, in _get_user
    user = msg['msg']['commit'].get('author', {}).get('login', {})
AttributeError: 'NoneType' object has no attribute 'get'
```

PR #421, opened 2017-03-28
Update wording of Pagure messages to match Github messages

danc86:
The main aim is consistency between the Pagure and Github messages, but this also has the side effect of tidying up some nits in the wording of the Pagure messages:

* "Pull request" is two words, not hyphenated, and the identifier shouldn't run together with the phrase "pull request", so: `opened pull request #123` instead of `opened pull-request#123`
* Changed `project "fm-orchestrator"` to just `fm-orchestrator`, since the word `project` is superfluous. That makes it is a bit more compact and easier to scan in a list.
* Moved the pull request subject to the end, with the project before it. Again, makes it easier to read in a list of e-mail subjects.

The latter point is particularly important when some pull request submitters tend to write very long subjects, so for example this mail I received:

```
jkaluza opened pull-request#454: "Retry communication with PDC on RuntimeError exception and increase the timeout to 120 seconds" on project "fm-orchestrator"
```

would instead be:

```
jkaluza opened pull request #454 on fm-orchestrator: Retry communication with PDC on RuntimeError exception and increase the timeout to 120 seconds
```

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/pagure.py
+++ b/fedmsg_meta_fedora_infrastructure/pagure.py
@@ -353,22 +353,20 @@ class PagureProcessor(BaseProcessor):
         elif 'pagure.project.forked' in msg['topic']:
             old_project = msg['msg']['project']['parent']['name']
             tmpl = self._(
-                '{user} forked project "{old_project}" to "{project}"'
+                '{user} forked {old_project} to {project}'
             )
             return tmpl.format(
                 user=user, old_project=old_project, project=project)
         elif 'pagure.pull-request.comment.added' in msg['topic']:
             prid = msg['msg']['pullrequest']['id']
             tmpl = self._(
-                '{user} commented on pull-request#{id} of project '
-                '"{project}"'
+                '{user} commented on PR #{id} on {project}'
             )
             return tmpl.format(user=user, id=prid, project=project)
         elif 'pagure.pull-request.comment.edited' in msg['topic']:
             prid = msg['msg']['pullrequest']['id']
             tmpl = self._(
-                '{user} edited a comment on pull-request#{id} of project '
-                '"{project}"'
+                '{user} edited a comment on PR #{id} on {project}'
             )
             return tmpl.format(user=user, id=prid, project=project)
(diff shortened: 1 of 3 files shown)


danc86:
@pypingou how would you feel about these wording changes for Pagure notifications?

fedmsg_meta_fedora_infrastructure/pagure.py
```diff
@@ -353,44 +353,40 @@ def subtitle(self, msg, **config):
...
-                '"{project}"'
+                '{user} commented on PR #{id} on {project}'
             )
             return tmpl.format(user=user, id=prid, project=project)
         elif 'pagure.pull-request.comment.edited' in msg['topic']:
             prid = msg['msg']['pullrequest']['id']
             tmpl = self._(
-                '{user} edited a comment on pull-request#{id} of project '
-                '"{project}"'
+                '{user} edited a comment on PR #{id} on {project}'
```
pypingou: Some time we use PR some time pull request, should we try to be consistent?

pypingou:
Just one comment, otherwise 👍 for me

fedmsg_meta_fedora_infrastructure/pagure.py
danc86: That's a good point. It uses "pull request" everywhere except for the messages which refer to comments, where it says "comment on PR". I just based it on the wording in `github.py` which uses that. I guess it is just to save a little bit of space.

I could update it to "comment on pull request", or we could even use "PR" instead of "pull request" everywhere as that is shorter... in that case I would want to update the Github messages to match. Tell me what you prefer. :-)

pypingou: I am quite neutral on this, go with what you prefer :)

danc86: In that case I would just leave it as it is here :-) whoever wrote the messages for Github struck a good balance I think so it makes sense to me to just follow the same patterns in the Pagure ones.

pypingou:
Alright, let's merge then :)

PR #423, opened 2017-03-28
Namespace support for pkgdb messages.

### LINKED ISSUE #422: Need to add namespace info to subtitle and link for pkgdb messages
I created a bunch of `modules/` repos but it sent notifications to all the package maintainers making it look like I was adding people to their repos that I shouldn't have.

Clarity in the messages would reduce alarm.

ralphbean:
Fixes #422.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/pkgdb.py
+++ b/fedmsg_meta_fedora_infrastructure/pkgdb.py
@@ -56,10 +56,11 @@ class PkgdbProcessor(BaseProcessor):
         if 'pkgdb.acl.update' in msg['topic']:
             tmpl = self._(
                 u"{agent} changed {user}'s '{acl}' permission " +
-                "on {package} ({branch}) to '{status}'"
+                "on {namespace}/{package} ({branch}) to '{status}'"
             )
             user = msg['msg']['username']
             status = msg['msg']['status']
+            namespace = msg['msg']['package_listing']['package'].get('namespace', 'rpms')
             package = msg['msg']['package_listing']['package']['name']
             acl = msg['msg']['acl']
             agent = get_agent(msg)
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/pkgdb.py
```diff
@@ -309,8 +338,10 @@ def subtitle(self, msg, **config):
 
             branch = _msg['action']['collection']['branchname']
             try:
+                namespace = _msg['action']['info'].get('namespace', 'rpms')
```
jeremycline: Same note about not needing to be inside the try/except with a default.

fedmsg_meta_fedora_infrastructure/pkgdb.py
```diff
@@ -611,11 +654,14 @@ def link(self, msg, **config):
             'package.branch.request',
         ])):
             try:
+                namespace = msg['msg']['package_listing']['package'].get('namespace', 'rpms')
```
jeremycline: Same note :smile:

jeremycline:
One other note, the ``packages`` function doesn't provide users with any namespace information. I'm not sure what sort of API stability guarantees there are for this stuff, but it's something that FMN needs in order to not message people who maintain, say, ``rpms/php`` for things that happen in ``modules/php``. 

If we can't change the API there's probably something we can do on FMN end to get this information, but it'd one more thing that breaks when message formats change, I guess.

For reference: https://github.com/fedora-infra/fmn/issues/176

ralphbean:
Thanks about the `msg2packages` observation.  I left some comments in fedora-infra/fmn#176 about it.

I'll take a look at the try/except issues today.

fedmsg_meta_fedora_infrastructure/pkgdb.py
```diff
@@ -106,87 +111,102 @@ def subtitle(self, msg, **config):
             agent = get_agent(msg)
 
             try:
+                namespace = msg['msg']['package_listing']['package'].get('namespace', 'rpms')
```
ralphbean: Ah, it will if `'package_listing'` is not in `msg['msg']`.  Look at the `except:` block and see that this is the case being handled by the original code.

fedmsg_meta_fedora_infrastructure/pkgdb.py
```diff
@@ -106,87 +111,102 @@ def subtitle(self, msg, **config):
...
                 package = msg['msg']['package']
 
             return tmpl.format(agent=agent, package=package, verb=verb,
-                               extra=extra)
+                               extra=extra, namespace=namespace)
         elif msg['topic'].endswith('pkgdb.package.update'):
             agent = get_agent(msg)
 
             try:
+                namespace = msg['msg']['package_listing']['package'].get('namespace', 'rpms')
```
ralphbean: Same response as above.  This is really a check for `'package_listing'`.

fedmsg_meta_fedora_infrastructure/pkgdb.py
ralphbean: Same response.  This is really a check to see if `'info'` is present.

fedmsg_meta_fedora_infrastructure/pkgdb.py
ralphbean: Ditto.  :p

fedmsg_meta_fedora_infrastructure/pkgdb.py
jeremycline: Ahhh that makes sense.

ralphbean:
Thanks @jeremycline!  (FYI, I don't have rights to merge, otherwise I would do it.)

jeremycline:
Oh weird, it shows you as a "Member" so I assumed you still had ACLs.

PR #424, opened 2017-03-29
Be a little less obvious when the assignee is missing

pypingou:


puiterwijk:
I think that instead of this, we should get the actual bug in Pagure looked at and hopefully fixed.

pypingou:
Yeah, I want to fix it in pagure itself but since we now have messages in datagrepper, we need to support them properly

fedmsg_meta_fedora_infrastructure/pagure.py
```diff
@@ -231,13 +231,13 @@ def subtitle(self, msg, **config):
...
-            else:
-                assignee = msg['msg']['issue']['assignee']['name']
+            assignee = ''
+            if msg['msg']['issue']['assignee']:
+                assignee = ' to {0}'.format(
+                    msg['msg']['issue']['assignee']['name']
+                )
             tmpl = self._(
-                '{user} assigned ticket {project}#{id} to {assignee}')
+                '{user} assigned ticket {project}#{id}{assignee}')
```
puiterwijk: How about: 
```
assignee = msg['msg']['issue'].get('assignee')
tmpl = self._(
    '{user} assigned ticket {project}#{id}' + (' to {assignee}' if assignee else ''))
```

pypingou: Sure

puiterwijk: Actually, you don't need the assignee variable in that case, so this would also suffice:

tmpl = self._(
    '{user} assigned ticket {project}#{id}' + (' to {assignee}' if ('assignee' in msg['msg']['issue']) else ''))

pypingou: Would work, but I find the first solution cleaner, or we use the construct we talked about yesterday:  ``assignee = (msg['msg']['issue'].get('assignee') or {}).get('name', '')``

puiterwijk: Right. I like the first one more too after rereading.

pypingou: Note that on the first we're missing the .get('name') part :)

pypingou: In a way, I find the current approach satisfactory :)

puiterwijk: Meh. I don't really like the .format to create a temporary string that you then .format into another one..

pypingou: Looks like we need a third opinion here, @bowlofeggs or @jeremycline want to chime in? :)

bowlofeggs: I don't have a strong opinion one way or the other. Both of the proposed approaches sound fine to me.

pypingou: Ok, so we need a fourth opinion, @jeremycline want to weigh in?

jeremycline: I don't really like burying an if/else statement inside a block that way.

pypingou: So what would you do?

jeremycline: Sorry I missed responding to this, I realize my comment wasn't very clear. I think it's better to not try to compress it into a one-liner at the expense of readability. However, you should do what you think is clearest since "don't really like" isn't a strong argument one way or the other.

pypingou: I kinda think the way it is currently is the easiest to read, the one-line approach is fun but not really readable.

pypingou:
Ok, we've been nitpicking about the style of the change but at the end no-one disagreed with the change in itself, so I'm going to make an arbitrary call here and just merge the PR.
If someone feels strong about it they can open a new PR and we can start bikesheding about this again :)

Thanks for your input everyone :)

PR #428, opened 2017-04-17
Add support for PR reviews in Github processor

sayanchowdhury:
Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/tests/github.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/github.py
@@ -214,6 +214,7 @@ class TestGithubPush(Base):
         }
     }
 
+
 class TestGithubIssueOpen(Base):
     """ There exists `a service
     <https://apps.fedoraproject.org/github2fedmsg>`_ to link the select github
@@ -330,6 +331,7 @@ class TestGithubIssueOpen(Base):
         }
     }
 
+
 class TestGithubIssueLabel(Base):
     """ There exists `a service
     <https://apps.fedoraproject.org/github2fedmsg>`_ to link the select github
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/tests/github.py
```diff
@@ -661,143 +664,1198 @@ class TestGithubIssueComment(Base):
...
                 "html_url": "https://github.com/fedora-infra/github2fedmsg",
                 "has_issues": True,
-                "master_branch": "develop",
-                "forks": 0,
-                "open_issues_count": 1,
-                "watchers": 2,
-                "name": "github2fedmsg",
+                "master_branch": "develop",
+                "forks": 0,
+                "open_issues_count": 1,
```
pypingou: Is this changing an existing test? If so, I don't think we should, add new one, deprecate the old one is fine, but we shouldn't change existing ones, imho.

pypingou:
The change to the processor looks fine but the changes to the tests are a little odd, it looks like some of the existing tests have been changed/edited. Is that the case?
We shouldn't change them since we need to maintain backward compatibility and the old messages are still present in datagrepper.

fedmsg_meta_fedora_infrastructure/tests/github.py
sayanchowdhury: Nope, I did not change any of the existing test-cases. Just added the new ones.

sayanchowdhury:
None of existing tests were altered. Though, I'll just go through the change once to make sure nothing has been removed.

sayanchowdhury:
Moved the tests to the end of the file. Looks more sane now.

pypingou:
Thanks, it's easier to read indeed, 👍 for me!

sayanchowdhury:
Thanks :+1:

PR #430, opened 2017-05-31
Deal with broken COPR messages

puiterwijk:
The "pkg" value suddenly switched from N-V-R to just name, with no
idea how to determine whether it is or not.

This patch is very obnoxious.
I want the COPR guys to get it fixed and consider their changes to
message format vs fedmsg-meta.


puiterwijk:
@xsuchy @clime Might be great if you'd fix it in COPR, and make sure your versions don't break with fedmsg-meta in the future. Thanks.

puiterwijk:
Breaking message: https://apps.fedoraproject.org/datagrepper/id?id=2017-931a61a3-42d8-4972-86b4-2431252c6010&is_raw=true&size=extra-large

PR #433, opened 2017-08-01
Turns out that pungi has an icon.  Who knew?

ralphbean:


pypingou:
I certainly didn't, 👍 as well!

Issue #434, opened 2017-08-01
The SCM module isn't namespace aware


jeremycline:
As reported in https://pagure.io/fedora-infrastructure/issue/6089,  the links being generated don't take the namespace into account.

PR #436, opened 2017-08-07
Adjust the scm links to the deployment of pagure over dist-git

pypingou:
While doing so fix retrieving the entire patch of a commit from pagure
as we used to do it from cgit.

Fixes https://pagure.io/fedora-infrastructure/issue/6204

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/scm.py
+++ b/fedmsg_meta_fedora_infrastructure/scm.py
@@ -1,5 +1,5 @@
 # This file is part of fedmsg.
-# Copyright (C) 2012-2014 Red Hat, Inc.
+# Copyright (C) 2012-2017 Red Hat, Inc.
 #
 # fedmsg is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -16,6 +16,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 #
 # Authors:  Ralph Bean <rbean@redhat.com>
+#           Pierre-Yves Chibon <pingou@pingoured.fr>
 #
 from fedmsg_meta_fedora_infrastructure import BaseProcessor
 from fedmsg_meta_fedora_infrastructure.fasshim import \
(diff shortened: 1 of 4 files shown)


fedmsg_meta_fedora_infrastructure/scm.py
```diff
@@ -56,15 +57,17 @@ def long_form(self, msg, **config):
             except KeyError:
                 repo = '.'.join(msg['topic'].split('.')[5:-1])
 
+            namespace = msg['msg']['commit'].get('namespace', 'rpms')
```
jeremycline: Nice, I think this will also take care of #434. Can you amend the commit message to note that you're fixing that issue, too?

fedmsg_meta_fedora_infrastructure/scm.py
```diff
@@ -56,15 +57,17 @@ def long_form(self, msg, **config):
...
             seen = msg['msg']['commit'].get('seen', False)
             if seen:
                 return already_seen_msg
 
-            url = 'https://src.fedoraproject.org/cgit/' + \
-                '{repo}.git/patch/?id={rev}'
-            response = requests.get(url.format(repo=repo, rev=rev))
+            url = 'https://src.fedoraproject.org/' + \
+                '{namespace}/{repo}/c/{rev}.patch'
+            response = requests.get(url.format(
```
jeremycline: Can you add a timeout to this? ``requests.get(..., timeout=30)`` or something. 

You'll want to catch all requests exceptions here as well since this API doesn't include raising any requests exceptions. ``requests.exceptions.RequestException`` is the general requests exception. I guess you could also call ``response.raise_for_status()`` to handle non-"okay" responses and then in the except block return a reasonable default explaining why the patch isn't included, otherwise return response.text.

fedmsg_meta_fedora_infrastructure/scm.py
pypingou: Sure will do

fedmsg_meta_fedora_infrastructure/scm.py
pypingou: Block changed to:
~~~~
            url = 'https://src.fedoraproject.org/' + \
                '{namespace}/{repo}/c/{rev}.patch'
            url = url.format(namespace=namespace, repo=repo, rev=rev)

            try:
                response = requests.get(url, timeout=30)
                if response.status_code == 200:
                    return response.text
            except requests.exceptions.RequestException:
                return "Could not successfully call: {}".format(url)
~~~~

pypingou:
Thanks for your review, I will merge as is so we can cut a new release if we want to improve the existing code it should be simple enough to iterate on this :)

PR #438, opened 2017-08-11
Deal with broken COPR messages

puiterwijk:
The "pkg" value suddenly switched from N-V-R to just name, with no
idea how to determine whether it is or not.

This patch is very obnoxious.
I want the COPR guys to get it fixed and consider their changes to
message format vs fedmsg-meta.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/coprs.py
+++ b/fedmsg_meta_fedora_infrastructure/coprs.py
@@ -84,8 +84,13 @@ class CoprsProcessor(BaseProcessor):
             # Zero pad buildid
             kwargs['build'] = '%08d' % (kwargs['build'])
 
-            # Also, extract the name from the NVR.
-            name, version, release = kwargs['pkg'].rsplit('-', 2)
+            if '-' not in kwargs['version']:
+                # This has happened.
+                # Example: 2017-c1e242e3-3a1c-4de9-8d37-61640b5e99fa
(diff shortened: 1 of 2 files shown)


puiterwijk:
This was PR #430 into master as it was an issue fix. Seemingly fedmsg-meta doesn't follow gitflow, so this is the same patch rebased for develop.

puiterwijk:
Note: we had been told that this had been a one-time bug in the COPR server, but seemingly the bug returned: https://apps.fedoraproject.org/datagrepper/id?id=2017-199ef77d-cdba-4f14-aaab-87ea4f6abb70&is_raw=true&size=extra-large

fedmsg_meta_fedora_infrastructure/coprs.py
```diff
@@ -84,8 +84,13 @@ def long_form(self, msg, **config):
             # Zero pad buildid
             kwargs['build'] = '%08d' % (kwargs['build'])
 
-            # Also, extract the name from the NVR.
-            name, version, release = kwargs['pkg'].rsplit('-', 2)
+            if '-' not in kwargs['version']:
+                # This has happened.
+                # Example: 2017-c1e242e3-3a1c-4de9-8d37-61640b5e99fa
+                name = '?? WHO KNOWS ??'
```
bowlofeggs: haha

PR #443, opened 2017-08-28
Handle pungi.compose.fail.to.start messages.

ralphbean:
/cc @puiterwijk.  This is the one you dug up in the FMN logs.


pypingou:
Thanks! :)

PR #444, opened 2017-08-31
Fix the compose2 processor when the compose messages has no compose_id

pypingou:


pypingou:
Would be nice to have the messages to add the test for this but that's the hotfix I put on notifs-backend01 to get it running

fedmsg_meta_fedora_infrastructure/compose2.py
```diff
@@ -80,5 +80,5 @@ def objects(self, msg, **config):
         elif 'file' in msg['msg']:
             return set([msg['msg']['file'].strip('/')])
         else:
-            compose = msg['msg']['compose_id']
+            compose = msg['msg'].get('compose_id', '')
```
bowlofeggs: I believe #443 also contains this patch, except it leaves off the ```''``` and so defaults to ```None```. I also think testing it would be ideal, but LGTM otherwise.

pypingou:
Then I guess #443 superseeds this one

PR #447, opened 2017-09-25
Add support for fedmsg messages of pagure over dist-git

pypingou:
This commit also adds support for stg vs prod messages (so the links for
example are consistent) and adds a couple of tests as the messages should
be the same since the software is the same, it is assumed two tests are
enough to test the inheritance of the processor and stg vs prod.


pypingou:
Thanks for the review :)

PR #450, opened 2017-10-24
A first pass at greenwave handling.

ralphbean:
Will likely need to add more example messages over time.

(In particular, messages for compose/rawhide gating and Atomic CI changes.)

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/greenwave.py
+++ b/fedmsg_meta_fedora_infrastructure/greenwave.py
@@ -1,0 +1,76 @@
+# This file is part of fedmsg.
+# Copyright (C) 2017 Red Hat, Inc.
+#
+# fedmsg is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# fedmsg is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with fedmsg; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# Authors:  Ralph Bean <rbean@redhat.com>
+#
+
+from six.moves.urllib.parse import urlencode
+from fedmsg_meta_fedora_infrastructure import BaseProcessor
+
+
+class GreenwaveProcessor(BaseProcessor):
+    __name__ = "greenwave"
+    __description__ = "Greenwave Gating Service"
+    __link__ = "greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0/policies"
+    __docs__ = "https://fedoraproject.org/wiki/Infrastructure/Factory2/Focus/Greenwave"
+    __obj__ = "Gating Decisions"
+    __icon__ = "https://apps.fedoraproject.org/img/icons/greenwave.png"
+
+    @staticmethod
+    def satisfied(msg):
+        try:
+            # https://pagure.io/greenwave/pull-request/100
+            return msg['msg']['policies_satisfied']
+        except KeyError:
+            return msg['msg']['policies_satisified']
+
+    def link(self, msg, **config):
+        subject = msg['msg']['subject']
+        if subject:
+            base = "https://taskotron.fedoraproject.org/resultsdb/results"
+            query = urlencode(msg['msg']['subject'][0])
+            return base + "?" + query
+
+    def subtitle(self, msg, **config):
+        if self.satisfied(msg):
+            decision = self._("is a GO")
+        else:
+            decision = self._("says NO-GO")
+
+        tmpl = self._(
+            "greenwave {decision} on {item} for "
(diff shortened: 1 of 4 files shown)


ralphbean:
See also https://apps.fedoraproject.org/datagrepper/raw?category=greenwave&delta=127800&rows_per_page=1

matt8754:
Looks good to me. However, I have no idea why the tests failed. :-)

ralphbean:
See also #452.

ralphbean:
Rebase on top of #451 completed.

ralphbean:
Any fedora-infra reviewers/mergers for this one?

fedmsg_meta_fedora_infrastructure/greenwave.py
```diff
@@ -0,0 +1,83 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with fedmsg; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# Authors:  Ralph Bean <rbean@redhat.com>
+#
+
+try:
```
jeremycline: I recommend using six instead of handling this yourself.

fedmsg_meta_fedora_infrastructure/greenwave.py
```diff
@@ -0,0 +1,83 @@
...
+from fedmsg_meta_fedora_infrastructure import BaseProcessor
+
+
+class GreenwaveProcessor(BaseProcessor):
+    __name__ = "greenwave"
+    __description__ = "Greenwave Gating Service"
+    __link__ = "greenwave-web-greenwave.app.os.fedoraproject.org/api/v1.0/policies"
+    __docs__ = "https://fedoraproject.org/wiki/Infrastructure/Factory2/Focus/Greenwave"
+    __obj__ = "Gating Decisions"
+    # XXX - We have a greenwave icon in the works.  Let's use this for now.
```
jeremycline: In my opinion it's better to write an issue for this sort of thing rather than a TODO

fedmsg_meta_fedora_infrastructure/greenwave.py
```diff
@@ -0,0 +1,83 @@
...
+    __icon__ = "https://apps.fedoraproject.org/img/icons/taskotron.png"
+
+    @staticmethod
+    def satisfied(msg):
```
jeremycline: I recommend documenting this method. Do you intend for a KeyError to be raised if the message is malformed? That doesn't seem like the best exception type for that, but it should be documented.

This applies to all the methods below, too.

fedmsg_meta_fedora_infrastructure/greenwave.py
```diff
@@ -0,0 +1,83 @@
...
+    def link(self, msg, **config):
+        subject = msg['msg']['subject']
+        if subject:
+            base = "https://taskotron.fedoraproject.org/resultsdb/results"
+            query = urlencode(msg['msg']['subject'][0])
+            return base + "?" + query
+
+    def subtitle(self, msg, **config):
+        if self.satisfied(msg):
+            decision = self._("is a GO")
```
jeremycline: You should handle the KeyError that can be raised in ``satisfied`` here, or handle it inside ``satisfied`` itself.

ralphbean:
> I really recommend more robust error handling in all the code. fedmsg.meta constantly raises unexpected exceptions due to poor error handling/undocumented APIs. Someone will alter the message schema in the future and this will fall over when that happens.

@jeremycline - how about the solution in https://github.com/fedora-infra/fedmsg/pull/493 instead?

pypingou:
Looks like this is now conflicting :)

ralphbean:
Thanks @pypingou.  Rebased!

pypingou:
Looks good to me

@jeremycline thoughts?

ralphbean:
Now with an icon.

jeremycline:
My comment about six stands, but I have no strong feelings one way or the other. fedmsg would need a release containing that change, as well, but I suppose it's no worse off than the rest of fedmsg_meta. Merge whenever you're ready.

fedmsg_meta_fedora_infrastructure/greenwave.py
```diff
@@ -0,0 +1,82 @@
...
+        else:
+            decision = self._("says NO-GO")
+
+        tmpl = self._(
```
pypingou: Note that this would probably make it the only processor using ``self._()``

fedmsg_meta_fedora_infrastructure/greenwave.py
pypingou: We're already using six elsewhere in the project, so using it here would also make sense

ralphbean: Ah, `six.moves.urllib.parse.urlencode` seems to have different behavior on py36 and py27.  See the test failure now?

fedmsg_meta_fedora_infrastructure/greenwave.py
ralphbean: I don't understand @pypingou.  What do you mean?

ralphbean:
Should be good-to-go with `six` now.

fedmsg_meta_fedora_infrastructure/greenwave.py
pypingou: Something entirely wrong apparently, nevermind me, should have checked before :)

pypingou:
Looks good to me :)

I'm hoping to make a release tomorrow, so it'll be in! :)

PR #451, opened 2017-11-07
Target py3.6

ralphbean:
We don't have any py2.6 targets anymore, and py3.4 has already flown by.

PR #452, opened 2017-11-07
WaiverDB processor.

ralphbean:
See also #450.


bowlofeggs:
It seems like rebasing this PR on top of #451 would help its tests to pass.

fedmsg_meta_fedora_infrastructure/tests/__init__.py
```diff
@@ -66,6 +66,7 @@
 from fedmsg_meta_fedora_infrastructure.tests.pdc import *
 from fedmsg_meta_fedora_infrastructure.tests.mbs import *
 from fedmsg_meta_fedora_infrastructure.tests.centos_ci import *
+from fedmsg_meta_fedora_infrastructure.tests.waiverdb import *
```
bowlofeggs: I recommend against this style of importing, as it's easy to end up with clashes, and even outside of clashes you can end up with a very polluted namespace. Explicit is better than implicit.

ralphbean:
Rebase on top of #451 completed.

ralphbean:
Can anyone help merge this one?

pypingou:
Let's merge :)

PR #453, opened 2017-11-27
Add support for the new pagure messages

pypingou:


pypingou:
alright, now the failures are no longer related to my changes :)

fedmsg_meta_fedora_infrastructure/pagure.py
```diff
@@ -241,6 +246,22 @@ def subtitle(self, msg, **config):
...
+            return tmpl.format(
+                user=user, project=project, id=issueid, tags=tags)
+        elif 'pagure.pull-request.tag.removed' in msg['topic']:
+            issueid = msg['msg']['pull_request']['id']
+            tags = msg['msg']['tags']
+            tags = fedmsg.meta.base.BaseConglomerator.list_to_series(tags)
+            tmpl = self._(
+                '{user} removed the {tags} tags from pull-request {project}#{id}')
+            return tmpl.format(
+                user=user, project=project, id=issueid, tags=tags)
```
bowlofeggs: Rather than these long strings of elif's, why not make a dict to map topics to small functions that handle them?

pypingou: That'd be one way to do it but that would mean refactoring all the code in this file and even to the entire project if one wants to be consistent in style.
So I don't think we'll want to do this :)

Plus I'm not entirely sure if the gain would be worth it :)

pypingou:
Thanks for the review!

PR #455, opened 2017-12-09
Substantial rewrite + extension of compose.* tests

AdamWill:
The tests for the 'compose' message topic were missing rather
a lot of coverage. This commit substantially revises and
extends the tests such that they should hopefully cover just
about all message forms that have existed. This commit does not
change the behaviour of the processor at all, and so some of the
expected values are sort of 'wrong' - they represent what the
processor currently does, even though the newly-added test makes
it clear the processor should probably do something else.
Subsequent commits will improve the processor behaviour, and of
course update the tests where appropriate.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/tests/compose.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/compose.py
@@ -23,17 +23,20 @@ from fedmsg.tests.test_meta import Base
 
 from .common import add_doc
 
+## LEGACY 'BRANCHED' MESSAGES
 
-class TestLegacyComposeBranchedComplete(Base):
-    """ This tests "old school" compose messages.
+
+class TestLegacyPreArchComposeBranchedComplete(Base):
+    """ This tests Branched compose complete messages from before the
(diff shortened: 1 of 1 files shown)


AdamWill:
Note, there are also some corrections to existing tests. e.g. the changes to the `branch` values - I checked old messages on datagrepper and the value really has always been just the number, i.e. `18` not `f18`.

AdamWill:
@ralphbean - I would also tag mohan, but I can't find his github ID, if he has one.

AdamWill:
Note, because this is such a big rewrite but all the test methods look quite similar, the diff is fairly nonsensical and impossible to read. You're better off just reading old and new copies of the file side-by-side to review this, I think.

mohanboddu:
LGTM, :+1:

AdamWill:
@ralphbean are you *sure* you can't merge this? I see many commits from you in the log. if not, @pypingou , could you do the honors? Thanks :)

pypingou:
Thanks! :)

PS: Ralph removed himself from the group and thus lost the merge abilities :(

PR #457, opened 2017-12-13
Fix `nodoc` to exclude test class from topic list

AdamWill:
It's been documented for a long time that you can set the class
attribute `nodoc` to `True` for a test class, to exclude that
class from appearing in the `topics.rst` file produced by
`make_topics_doc`, which is publicly published as the definitive
list of message topics. However, this has never actually worked
properly, and test classes with `nodoc = True` *do* appear in
the list. This fixes it, by checking the attribute correctly.


AdamWill:
@ralphbean @pypingou

pypingou:
Cool, thanks!

PR #459, opened 2017-12-14
Revise subtitles for compose.* messages

AdamWill:
The current subtitles don't provide as much information as they
could, sometimes use unhelpful jargon, hide ambiguity in one
case, and occasionally are entirely wrong (some message topics
aren't specifically covered, which leads to them being given
subtitles as if they were the overall compose start/stop
messages, which they aren't). This commit revises the subtitle
handling for the compose.* processor to address these issues,
and updates the tests accordingly, including adding tests for
the missing topics ('image.start' and 'image.complete').

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/compose.py
+++ b/fedmsg_meta_fedora_infrastructure/compose.py
@@ -27,88 +27,125 @@ class ComposeProcessor(BaseProcessor):
     __docs__ = "https://pagure.io/docs/releng"
     __obj__ = "Composes"
 
+    # dicts mapping topic substrings to subtitle event texts
+    # These are the old 'make-updates' topics
+    makeupevents = {
+        '.make-updates.':       "",
+        '.cloudimg-build.':     " building images for",
+        '.mash-atomic':         " Atomic updates mash for",
+        '.atomic-lorax':        " Atomic lorax run for",
+        '.cloudimg-checksum.':  " image checksum generation for",
+        '.cloudimg-staging.':   " master mirror publication for",
+    }
+    # as these events only existed prior to pungi 4 and the creation
+    # of different 'dists', we can always treat the shortname / dist
+    # as 'Fedora' for these
+    oldevents = {
+        '.mash.':               " mashing",
+        '.pungify.':            " building boot.iso for",
+        '.image.':              " building other images for",
+    }
+    oldevents.update(makeupevents)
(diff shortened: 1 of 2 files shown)


AdamWill:
@pypingou @mohanboddu

fedmsg_meta_fedora_infrastructure/compose.py
```diff
@@ -27,88 +27,125 @@ class ComposeProcessor(BaseProcessor):
     __docs__ = "https://pagure.io/docs/releng"
     __obj__ = "Composes"
 
+    # dicts mapping topic substrings to subtitle event texts
+    # These are the old 'make-updates' topics
+    makeupevents = {
+        '.make-updates.':       "",
```
pypingou: For these dictionaries, do we have anything that'd be a little more specific, maybe ``releng.make-updates.`` & so on? I fear (down the road) a little the conflicts on topics if we stay that generic

fedmsg_meta_fedora_infrastructure/compose.py
```diff
@@ -27,88 +27,125 @@ class ComposeProcessor(BaseProcessor):
...
+    allevents = dict()
+    allevents.update(oldevents)
+    allevents.update(events)
+
     def subtitle(self, msg, **config):
 
+        # decide what event we're dealing with
+        gotevent = False
+        for event in self.allevents:
+            if event in msg['topic']:
```
pypingou: The old structure seemed to rely on ``.endswith()`` most of the time, that makes the topic a little more specific (cf my concern above)

pypingou:
A couple of comments but overall, looks fine to me

fedmsg_meta_fedora_infrastructure/compose.py
AdamWill: AIUI, the messages each processor operates on are determined by its `__name__` property. This processor will only ever be used for `compose.*` messages. Let me know if this is wrong.

fedmsg_meta_fedora_infrastructure/compose.py
AdamWill: I didn't like the `endswith` because it meant there had to be two of everything, one for 'start' and one for 'complete'. This way seems cleaner, for me.

pypingou:
Alright, let's merge :)

PR #462, opened 2018-01-25
compose: just get the 'link' value from the fedmsg

AdamWill:
This big pile of mess that tries to figure out the correct URL
for a compose from the other values in the fedmsg is ugly as
heck and doesn't work right; it hasn't been updated for the
several compose types that have been introduced in the last few
years. It can even cause crashes on some fedmsgs. So let's just
throw it away entirely and go for a different approach.

https://pagure.io/pungi-fedora/pull-request/503 enhances the
compose scripts to emit a 'location' key with the correct URL
for the compose. So for future fedmsgs once that is merged, we
can just take that URL and use it. Simple! For older fedmsgs
we won't bother trying to do anything smart, we'll just return
the base location for Pungi 4 composes. There's no value really
to trying to 'do the right thing' for older messages as the
composes they refer to won't be available for that long anyway;
no-one needs to know where a compose from two years ago that was
deleted 1 year, 11 months and 2 weeks ago lived.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/compose.py
+++ b/fedmsg_meta_fedora_infrastructure/compose.py
@@ -148,32 +148,7 @@ class ComposeProcessor(BaseProcessor):
                            arch=arch, comptext=comptext)
 
     def link(self, msg, **config):
-        branch = msg['msg'].get('branch', 'not-a-number')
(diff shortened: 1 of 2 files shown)


AdamWill:
@puiterwijk

PR #466, opened 2018-02-22
Add Fedora Hubs

abompard:
Add handling of the fedmsg produced by Fedora Hubs.

There's also a couple conglomerate functions in case user change their hub or widget configuration repeatedly in a short time.


pypingou:
Since Fedora Hubs has been stopped for now, let's close this PR.

Sorry for this @abompard

PR #467, opened 2018-03-14
Fix the URL on messages about edited comment in pagure

pypingou:


pypingou:
Thanks folks! :)

PR #470, opened 2018-03-26
Generalize the CI pipeline processor and add support for the Allpackage messages

pypingou:


pypingou:
Thanks for the review!

PR #476, opened 2018-04-10
Merge the Allpackages processor into the Atomic CI processor

pypingou:
By itself this is not ideal, we are using the same processor for messages coming
from two different CI pipelines running in the ci.centos.org infrastructure.
However, we do not have really the choice, the first pipeline deployed there
is sending org.centos.prod/stage.ci.pipeline.<action> while the allpackages
pipeline is now sending messages of the type:
org.centos.prod/stage.ci.pipeline.allpackages and this is caught by the
Atomic CI processor.

So we had two ways to fix this:
- adjust the topic sent by the Atomic CI pipeline to include something like
  atomic_ci where the other pipeline is using 'allpackages'
- hack the processor to support both the atomic_ci and the allpackages format

Since we aim at being backward compatible, the second step was needed anyway
to support old and potentially new format of the messages sent by the atomic_ci
pipeline.
So since we do the second fix anyway, it is just as easy to not adjust the topic
sent by the Atomic CI pipeline.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/centos_ci.py
+++ b/fedmsg_meta_fedora_infrastructure/centos_ci.py
@@ -54,120 +54,127 @@ class AtomicCiProcessor(BaseProcessor):
                 status = 'failed'
             return status
 
-        if '%s.package.ignore' % self.__name__ in msg['topic']:
+        name = self.__name__
+        pipeline_name = self.pipeline_name
+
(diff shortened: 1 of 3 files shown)


bstinsonmhk:
I'm ok to merge now

pypingou:
thanks :)

PR #486, opened 2018-05-24
Fixing typo and Greenwave msg link params order

gnaponie:
Fixing "policies_satisified" typo.
And adding some code to be sure that the parameters in the link in
the Greenwave message will always have the same sorting so that
the test will always pass.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/greenwave.py
+++ b/fedmsg_meta_fedora_infrastructure/greenwave.py
@@ -32,17 +32,17 @@ class GreenwaveProcessor(BaseProcessor):
 
     @staticmethod
     def satisfied(msg):
-        try:
-            # https://pagure.io/greenwave/pull-request/100
-            return msg['msg']['policies_satisfied']
-        except KeyError:
-            return msg['msg']['policies_satisified']
+        return msg['msg']['policies_satisfied']
 
     def link(self, msg, **config):
         subject = msg['msg']['subject']
         if subject:
             base = "https://taskotron.fedoraproject.org/resultsdb/results"
-            query = urlencode(msg['msg']['subject'][0])
+            # we want to be sure that we always have the same parameter order
+            query = urlencode(sorted(
+                [(k, v) for k, v in msg['msg']['subject'][0].items()],
+                key=lambda tup: tup[0]
+            ))
             return base + "?" + query
 
     def subtitle(self, msg, **config):
--- a/fedmsg_meta_fedora_infrastructure/tests/greenwave.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/greenwave.py
@@ -52,7 +52,7 @@ class TestGreenwaveDecisionChange(Base):
         "crypto": "x509",
         "topic": "org.fedoraproject.prod.greenwave.decision.update",
         "msg": {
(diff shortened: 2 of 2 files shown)


fedmsg_meta_fedora_infrastructure/greenwave.py
```diff
@@ -32,17 +32,17 @@ class GreenwaveProcessor(BaseProcessor):
 
     @staticmethod
     def satisfied(msg):
-        try:
-            # https://pagure.io/greenwave/pull-request/100
```
puiterwijk: We can't remove this, since this code needs to remain working with all previously sent messages.

fedmsg_meta_fedora_infrastructure/tests/greenwave.py
```diff
@@ -52,7 +52,7 @@ class TestGreenwaveDecisionChange(Base):
         "crypto": "x509",
         "topic": "org.fedoraproject.prod.greenwave.decision.update",
         "msg": {
-            "policies_satisified": True,
+            "policies_satisfied": True,
```
puiterwijk: We'll need to keep the previous test cases and just add new tests for the fixed versions.

gnaponie: ACK! So I'll rebase with only the other change. Thanks

fedmsg_meta_fedora_infrastructure/greenwave.py
```diff
@@ -32,17 +32,17 @@ class GreenwaveProcessor(BaseProcessor):
...
+        return msg['msg']['policies_satisfied']
 
     def link(self, msg, **config):
         subject = msg['msg']['subject']
         if subject:
             base = "https://taskotron.fedoraproject.org/resultsdb/results"
-            query = urlencode(msg['msg']['subject'][0])
+            # we want to be sure that we always have the same parameter order
+            query = urlencode(sorted(
+                [(k, v) for k, v in msg['msg']['subject'][0].items()],
```
puiterwijk: I have been told that the `subject` might be more complicated then a simple key-value.
So we'll need to make sure that deeper nested subjects are also correctly encoded probably.

ralphbean: Really?

puiterwijk: @ralphbean I can't find the exact quote, but it was strongly alluded to that that was one of the main reasons none of my alternative suggestions could be used in https://pagure.io/waiverdb/issue/134#comment-496071 .

I think I remember having a discussion on IRC later about that where people told me that the "subject" for things can be pretty much anything and that "subjects from the past don't give guarantees for the future", and that that's why subjects need to be encoded as json throughout the stack.

Unless waiverdb and greenwave both have different views on what a "subject" is?

fedmsg_meta_fedora_infrastructure/greenwave.py
ralphbean: Yeah, agreed.

pypingou:
This doesn't seem to be ready to be merged :(

gnaponie:
I thought we dropped this one. I guess we can now.

PR #487, opened 2018-05-28
Add support for fedmsg message from pagure about PR assignee

pypingou:
Fixes https://pagure.io/pagure/issue/1955


pypingou:
Thanks for the review :)

PR #488, opened 2018-06-05
Update releng.atomic.twoweek fedmsg doc with multi-arch content

sinnykumari:
We have updated releng push-two-week-atomic script to
send multi-arch information as well in fedmsg.
Link - https://pagure.io/releng/c/ff2f2a5e035fa41507553acbc4f2fa15cc1873f5?branch=master

Signed-off-by: Sinny Kumari <sinny@redhat.com>

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/tests/releng.py
+++ b/fedmsg_meta_fedora_infrastructure/tests/releng.py
@@ -29,10 +29,280 @@ from .common import add_doc
 class TestTwoWeekAtomicBegin(Base):
     """ As one of the final steps in the `two week atomic
     <https://fedoraproject.org/wiki/Changes/Two_Week_Atomic>`_ process, there
+    is a `release engineering script
+    <https://pagure.io/releng/blob/master/f/scripts/push-two-week-atomic.py>`_ that
+    runs to perform the release.
+
+    During the two week script run, it fetches AtomicHost artifacts detail
+    for given pungi compose ID using `fedfind
+    <https://pagure.io/fedora-qa/fedfind/>`_ . After getting image
+    artifacts and ostree commit information successfully for all supported
+    architectures, it sends this fedmsg indicating that Two Week release process has started.
+    """
+
+    expected_title = "releng.atomic.twoweek.begin"
+    expected_subti = "Release engineering scripts started evaluating a " + \
+        "new set of builds for a Fedora 28 Atomic Host release"
+    expected_icon = "https://apps.fedoraproject.org/img/icons/atomic.png"
+    expected_secondary_icon = "https://apps.fedoraproject.org/img/icons/atomic.png"
+    expected_link = "https://download.fedoraproject.org/pub/alt/atomic/stable/"
+    expected_objects = set([u'aarch64', u'x86_64', u'ppc64le'])
+
+    msg = {
+        u'timestamp': 1444939709.0,
+        u'topic': u'org.fedoraproject.prod.releng.atomic.twoweek.begin',
+        u'msg': {
+            u'aarch64': {
+                u'atomic_dvd_ostree': {
+                    u'name': u'Fedora-AtomicHost-ostree-aarch64-28-20180515.1.iso',
+                    u'image_name': u'Fedora-AtomicHost-ostree-aarch64-28-20180515.1.iso',
+                    u'image_url': u'/pub/alt/atomic/stable/Fedora-Atomic-28-20180515.1/AtomicHost/'
+                    'aarch64/iso/Fedora-AtomicHost-ostree-aarch64-28-20180515.1.iso',
+                    u'release': u'28',
+                    u'compose_id': u'Fedora-Atomic-28-20180515.1',
+                    u'size': 988649472
+                },
+                u'atomic_qcow2': {
+                    u'name': u'Fedora-AtomicHost-28-20180515.1',
+                    u'image_name': u'Fedora-AtomicHost-28-20180515.1.aarch64.qcow2',
+                    u'image_url': u'/pub/alt/atomic/stable/Fedora-Atomic-28-20180515.1/'
+                    'AtomicHost/aarch64/images/Fedora-AtomicHost-28-20180515.1.aarch64.qcow2',
+                    u'release': u'28',
+                    u'compose_id': u'Fedora-Atomic-28-20180515.1',
+                    u'size': 621911040
+                },
+                u'atomic_raw': {
+                    u'name': u'Fedora-AtomicHost-28-20180515.1',
+                    u'image_name': u'Fedora-AtomicHost-28-20180515.1.aarch64.raw.xz',
+                    u'image_url': u'/pub/alt/atomic/stable/Fedora-Atomic-28-20180515.1/'
+                    'AtomicHost/aarch64/images/Fedora-AtomicHost-28-20180515.1.aarch64.raw.xz',
+                    u'release': u'28',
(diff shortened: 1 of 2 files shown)


sinnykumari:
From the [CI log](https://travis-ci.org/fedora-infra/fedmsg_meta_fedora_infrastructure/jobs/388347228), error message says:
Installed /home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/arrow-0.12.1-py2.7.egg
error: urllib3 1.23 is installed but urllib3<1.23,>=1.21.1 is required by set(['requests'])

Error message looks unrelated to me with the change made in this PR.

fedmsg_meta_fedora_infrastructure/tests/releng.py
```diff
@@ -27,6 +27,187 @@
...
+    During the two week script run, it fetches AtomicHost artifacts detail
+    for given pungi compose ID using `fedfind
+    <https://pagure.io/fedora-qa/fedfind/>`_ . After getting image
+    artifacts and ostree commit information successfully for all supported
+    architectures, it sends this fedmsg indicating that Two Week release process has started.
+    """
+
+    expected_title = "releng.atomic.twoweek.begin"
+    expected_subti = "Release engineering scripts started evaluating a " + \
+        "new set of builds from Fedora 22 Atomic Host release"
```
dustymabe: 22? :)

fedmsg_meta_fedora_infrastructure/tests/releng.py
```diff
@@ -27,6 +27,187 @@
...
+    expected_secondary_icon = "https://apps.fedoraproject.org/img/icons/atomic.png"
+    expected_link = "https://download.fedoraproject.org/pub/alt/atomic/stable/"
+    expected_objects = set(['atomic_qcow2', 'atomic_raw'])
+
+    msg = {
+        u'timestamp': 1444939709.0,
+        u'topic': u'org.fedoraproject.prod.releng.atomic.twoweek.begin',
+        u'msg': {
+            u'aarch64': {
+                u'atomic_qcow2': {
```
dustymabe: I forget, did we rule out putting ISO image data in the fedmsg? right here we only have qcow and raw image data

fedmsg_meta_fedora_infrastructure/tests/releng.py
sinnykumari: Yeah, looks like it is true. I don't have much information of past on how it got started :)

fedmsg_meta_fedora_infrastructure/tests/releng.py
sinnykumari: No, I just added qcow2 and raw image for all arches to keep it a bit short. I am ok to add complete message which will include iso for all three arches and vagarant images detail for x86_64. Is that ok?

dustymabe: ahh so the real fedmsg does have the ISO in it?

dustymabe: yeah let's just make it match what the real fedmsg outputs. I think the fedmsg includes the ISO data, but if it doesn't we should probably add it.

sinnykumari: Existing fedmsg dosn't have ISO information. But we will have ISO information from next atomic twoweek release with recent change we did in push-two-week-atomic script. I will update this PR with exact content that we will have from our next atomic twoweek release fedmsg.

dustymabe: yes, thanks

sinnykumari:
Updated PR with full fedmsg content

dustymabe:
LGTM, but will want someone else that knows this code better to review.

sinnykumari:
@sayanchowdhury @pypingou Can you please review this PR?

pypingou:
When I run the tests with this PR I get 2 failures and 2 errors:

````
fedmsg/tests/test_meta.py", line 196, in _equals
    self.assertEquals(actual, expected)
AssertionError: Items in the second set but not the first:
'atomic_dvd_ostree'
'atomic_qcow2'
'atomic_raw'
'atomic_vagrant_libvirt'
'atomic_vagrant_virtualbox'
....
Failed at:
(' ', <fedmsg_meta_fedora_infrastructure.tests.TestTwoWeekAtomicComplete testMethod=test_objects>
````
and:
````
fedmsg_meta_fedora_infrastructure/releng.py", line 32, in subtitle
    release = msg['msg']['atomic_raw'].get('release', '')
KeyError: 'atomic_raw'
````

**Comment by sinnykumari** [comment]
Thanks @pypingou for reviewing it. Updated the change to fix test, it passes locally for me.

**Comment by sinnykumari** [comment]
Is this PR ok to get merged now?

**Comment by pypingou** [comment]
Nope, the previous approach was the correct one, existing tests should not be touched except to be marked ``Legacy`` as we still need to support old messages (present in datagrepper for example).
So you need to find a way to support both old and new messages.

**Comment by sinnykumari** [comment]
Updated PR, we have now test for both old and new messages. Test passes locally.

File: fedmsg_meta_fedora_infrastructure/tests/releng.py [hunk 3]
```diff
@@ -108,7 +376,6 @@ class TestTwoWeekAtomicComplete(Base):
                 "name": "Fedora-Cloud-Atomic-Raw"
             }
         }
-    }
-
+}
 
 add_doc(locals())
```
pypingou: The 2 empty lines at the end may be a pep8/flake8 requirement :)

pypingou:
Looks all good to me and tests are passing as well.

@sinnykumari after discussing with @puiterwijk about the merit of changing the topic vs changing the structure of the message, do you still want to continue with this PR?

fedmsg_meta_fedora_infrastructure/tests/releng.py
sinnykumari: Ah yes, fixed it :)

sinnykumari:
Summarizing here from the IRC conversation -
We have two option to proceed:

1\. Create a new fedmsg topic - suggestion from @puiterwijk:
Since new message's structure is not compatible with existing one, create a new fedmsg topic for new messages.

Advantage:
- Consumers who relies on the existing message's structure won't have to do any changes.

Disadvantage:
- Extra effort required since we will have to make changes in [fmn](https://github.com/fedora-infra/fmn) and [fedmsg_meta_fedora_infrastructure](https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure)
- Existing consumers like fedora-website will no longer get updated message's on existing fedmsg topic. They will have to start listening to new fedmsg topic.

2\. Continue using existing fedmsg topic for newer message:
Advantage:
- All changes are in place, so no extra effort required
- Consumers can continue listening to same fedmsg topic

Disadvantage:
 - Existing consumers like fedora-website will have to adapt to new message structure

Please correct me or add if I missed something.

Considering both option, I am more inclined to go with option 2 which is updating existing fedmsg topic with newer content.

dustymabe:
I vote for number 2 as well. Since we think the only consumer of this topic is the website. I can't think of any others.

pypingou:
> Existing consumers like fedora-website will have to adapt to new message structure

This is true for both options

sinnykumari:
@pypingou Can we merge this PR, so far there is no objection :)

sinnykumari:
We will need this commit deployed in prod before we do next Atomic Two Week release (which is most likely going to happen on Tuesday, 12th June).

PR #489, opened 2018-07-27
greenwave: Fix subtitle and result link for new messages

hluk:
---
E.g.: https://apps.fedoraproject.org/datagrepper/id?id=2018-3f0f22f8-6f94-4d8a-95c6-399ffdcf1ada&is_raw=true&size=extra-large


hluk:
Oh, I just realized that the result link will be still incorrect in some cases. This probably needs to be fixed in Greenwave. I'm reverting those changes.

hluk:
This PR fixes only message subtitle now.

For the other issue I've create [Greenwave bug report](https://pagure.io/greenwave/issue/268).

PR #491, opened 2018-10-12
Add container pipeline topics

johnbieren:
There is a new pipeline that is apples to apples with the all packages fedora pipeline, but for the container namespace. Because it is just on containers, there is only interesting stage in this pipeline (the functional tests). I hope to turn the messages on for these soon, so I am hoping to add them to fedmsg_meta.


pypingou:
Could you add tests for this please?

johnbieren:
Sure thing

johnbieren:
You were definitely right about that, thanks! Unfortunately, I am still failing. Perhaps I missed adding it to some other file as well?

johnbieren:
@pypingou Can we merge this?

pypingou:
@johnbieren I'll do that but I'd prefer to wait until I have a little time to cut a release that we can push out :)

Issue #495, opened 2019-03-19
CentOS pipeline messages sent with `null` namespace


puiterwijk:
e.g. https://apps.fedoraproject.org/datagrepper/id?id=2019-6aa280a3-05bf-48a1-affd-2e28ddf3e028&is_raw=true&size=extra-large

Note that fedmsg-meta will need to be fixed to be able to process these mesages, since we'll have these in datanommer forever.
The bug is in https://github.com/fedora-infra/fedmsg_meta_fedora_infrastructure/blob/develop/fedmsg_meta_fedora_infrastructure/centos_ci.py#L241

The full traceback is:
````
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]: [2019-03-18 12:52:05][fmn.formatters   ERROR] An unexpected error occurred while processing an IRC batch
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]: Traceback (most recent call last):
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]:   File "/usr/lib/python2.7/site-packages/fmn/formatters.py", line 185, in irc_batch
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]:     conglomerated = fedmsg.meta.conglomerate(messages, **config.app_conf)
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]:   File "/usr/lib/python2.7/site-packages/fedmsg/meta/__init__.py", line 207, in conglomerate
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]:     [message], subject=subject, lexers=lexers, **config)
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]:   File "/usr/lib/python2.7/site-packages/fedmsg/meta/base.py", line 328, in produce_template
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]:     }) for msg in constituents])
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]:   File "/usr/lib/python2.7/site-packages/fedmsg/meta/__init__.py", line 152, in __wrapper
Mar 18 12:52:05 notifs-backend01.phx2.fedoraproject.org celery[13462]:     return f(msg, **config)
[...]

Issue #484, opened 2018-05-23
Filtering of Fedora CI related messages broken


psss:
I've tried to configure email notifications for the CI pipeline but it seems the filtering is not working as expected. Here's what I tested:

**Single rule "All CI events"**

This shows recent results but those related to fedora-pr-pipeline are empty (no message description).

**"All CI events" + "A particular user"**

This finds messages but only about month old. The recent messages from the fedora-pr-pipeline are not displayed.

**"All CI events" + "A particular package"**

No messages found. Tested with "firefox" as the package which was listed in the "All CI events" output with the link pointing here:
https://jenkins-continuous-infra.apps.ci.centos.org/job/ci-pipeline-trigger/3855/

This is basically making the notifications related to CI unusable because it's not possible to filter messages releated to specific user/package. Could you please have a look into this? Thanks.

psss...

thrix:
after investigation I found out that the messages have changed again:

allpackages-build (is the all packages build pipeline now):
https://apps.fedoraproject.org/datagrepper/raw?topic=org.centos.prod.ci.pipeline.allpackages-build.package.complete

allpackages-pr (is the all packages pr pipeline now):
https://apps.fedoraproject.org/datagrepper/raw?topic=org.centos.prod.ci.pipeline.allpackages-pr.package.complete

We will need additional patches to cover this

thrix:
I will try to prepare a patch for both and submit it as a PR

thrix:
Filed an issue to get the docs fixed in the pipeline:

https://github.com/CentOS-PaaS-SIG/upstream-fedora-pipeline

PR #496, opened 2019-04-12
Fix CI pipeline processor for current messages (#495) (#484)

AdamWill:
The CI pipeline processor has an outright bug causing packagers
to be sent lots of error messages (#495), and aside from that
could not process most current messages, as it was not updated
for the introduction of separate 'allpackages-build' and
'allpackages-pr' messages (#484).

This fixes all cases where something we look at might be set to
None, and revises the tests heavily to test all current message
topics that I could find, and mark all old ones as legacy. The
tests use real messages with a couple of lengthy values taken
out and all the usernames changed to 'jankratochvil' as it makes
things easier.


AdamWill:
@pypingou @thrix @puiterwijk

PR #498, opened 2019-06-07
Bodhi 4 fix

AdamWill:
This mostly fixes the bodhi processor to handle some changes to messages in Bodhi 4. A few other fixes and test enhancements come in along the way. Note this does not, I think, make the processor fully compatible with just the *documented schema* for messages in Bodhi 4, which leaves out quite a lot of 'old' keys the processor uses. It only makes it compatible with the messages actually emitted by Bodhi 4.0, which still include a lot of the old keys for now.

@bowlofeggs @ralphbean @pypingou


AdamWill:
@puiterwijk too

PR #499, opened 2019-06-27
Bodhi schema compliance + more

AdamWill:
This branch started out aiming to make the processor respect the schemas Bodhi now publishes for messages - i.e. only to rely on things which are guaranteed to exist by the schemas. It mostly accomplishes that, plus fixes some other stuff I ran into on the way - messages whose contents have changed, a couple of new message topics, stuff like that. It also factors out some huge `elif` forests to make things a bit easier to follow, though there's not a lot I can do about `subtitle`.

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/bodhi.py
+++ b/fedmsg_meta_fedora_infrastructure/bodhi.py
@@ -43,7 +43,24 @@ def get_sync_product(msg):
     return product
 
 
+def get_sync_release(msg):
+    """Return the release for an ftpsync message, determining if we're
+    modular along the way. Returns the release and a boolean for "are
+    we modular"
+    """
+    modular = False
+    release = msg['msg']['release']
+    if release.endswith('m'):
+        modular = True
+        release = release.rstrip('m')
+    return (release, modular)
+
+
 def is_ftp_sync(msg):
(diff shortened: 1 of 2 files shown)


AdamWill:
@bowlofeggs @nirik @puiterwijk

fedmsg_meta_fedora_infrastructure/bodhi.py
```diff
@@ -43,7 +43,24 @@ def get_sync_product(msg):
     return product
 
 
+def get_sync_release(msg):
```
puiterwijk: Please note that we have more release postfix letters: `f` for flatpack, and `c` for container.

AdamWill:
@puiterwijk we could fix that, sure. Where do/will the flatpak and container update composes be synced to?

AdamWill:
Yeah, as far as I've seen, we don't get this 'release number plus type' string in any other messages, they indicate the same information in other ways. If it turns out that it does show up in other messages, though, we can rejig as a future PR.

AdamWill:
Further note: this 'release number plus type' string does show up in `request` messages, like `bodhi.update.request.testing`. However, the processor doesn't do anything with the string for those messages. I *think* the processor should actually act more or less as expected for those updates as-is, though it's a bit surprising.

PR #503, opened 2020-05-27
added ability to use fasjson when provided with config

StephenCoady:
This commit adds a function which allows the cache to be built using fasjson instead of fas.

there is a difference in how it does this. When using fas it used `python-fedora` to search using the alphabet and a wildcard `a*, b*` etc. Instead of a plain search `fasjson-client` exposes pagination functionality so I have used that.

Signed-off-by: Stephen Coady <scoady@redhat.com>

### DIFF
--- a/fedmsg_meta_fedora_infrastructure/fasshim.py
+++ b/fedmsg_meta_fedora_infrastructure/fasshim.py
@@ -108,7 +108,7 @@ def make_fasjson_cache(**config):
 
     creds = config['fas_credentials']
 
-    default_url = 'https://admin.fedoraproject.org/accounts/'
+    default_url = 'https://fasjson.fedoraproject.org/v1/'
 
     try:
         client = fasjson_client.Client(url=creds.get('base_url', default_url))
--- a/fedmsg_meta_fedora_infrastructure/fasshim.py
+++ b/fedmsg_meta_fedora_infrastructure/fasshim.py
@@ -106,7 +106,7 @@ def make_fasjson_cache(**config):
         log.warn("No fasjson-client installed.  Not caching fasjson.")
         raise
 
-    creds = config['fas_credentials']
+    creds = config['fasjson_credentials']
 
     default_url = 'https://fasjson.fedoraproject.org/v1/'
 
--- a/fedmsg_meta_fedora_infrastructure/fasshim.py
+++ b/fedmsg_meta_fedora_infrastructure/fasshim.py
@@ -1,5 +1,6 @@
 import collections
 import logging
+import os
 import socket
 import string
 import threading
@@ -100,26 +101,79 @@ def make_fasjson_cache(**config):
 
     log.warn("No previous fasjson cache found. Looking to rebuild.")
 
+    creds = config['fasjson_credentials']
+
+    krb5_principal = creds.get("krb5_principal")
+    krb5_client_ktname = creds.get("krb5_client_ktname")
+    gss_use_proxy = creds.get("gss_use_proxy")
+
+    if krb5_client_ktname:
+        os.environ["KRB5_CLIENT_KTNAME"] = krb5_client_ktname
+
+    if gss_use_proxy:
+        os.environ["GSS_USE_PROXY"] = "yes"
+
+    default_url = 'https://fasjson.fedoraproject.org/v1/'
+    base_url = creds.get('base_url', default_url)
+
     try:
         import fasjson_client
     except ImportError:
-        log.warn("No fasjson-client installed.  Not caching fasjson.")
-        raise
-
-    creds = config['fasjson_credentials']
-
-    default_url = 'https://fasjson.fedoraproject.org/v1/'
-
-    try:
-        client = fasjson_client.Client(url=creds.get('base_url', default_url))
-    except fasjson_client.errors.ClientSetupError as e:
-        log.error(
-            "Something went wrong setting up fasjson client with error: %s" % e
+        fasjson_client = None
+        log.warn(
+            "No fasjson-client installed.  Falling back to querying directly."
         )
-        return {}
+
+    if fasjson_client:
+        try:
+            client = fasjson_client.Client(
+                url=base_url, principal=krb5_principal
+            )
+        except fasjson_client.errors.ClientSetupError as e:
+            log.error(
+                "Error while setting up fasjson client: %s" % e
+            )
+            return {}
+        APIError = fasjson_client.errors.APIError
+    else:
+        import requests
+        import requests.exceptions
+        from requests.compat import urlencode, urljoin
+        from requests_gssapi import HTTPSPNEGOAuth
+
+        # shim inside a shim
+        class Client(object):
+            def __init__(self, url, principal=None):
+                self.url = url
+                self.principal = principal
+
+                gssapi_auth = HTTPSPNEGOAuth(
+                    opportunistic_auth=True, mutual_authentication="OPTIONAL"
+                )
+                self.session = requests.Session()
+                self.session.auth = gssapi_auth
+
+            def list_all_entities(self, ent_name):
+                if not ent_name.endswith("/"):
+                    # avoid redirection round trip
+                    ent_name += "/"
+                endpoint = urljoin(self.url, ent_name)
+
+                # yay, pagination
+                next_page_url = endpoint + "?" + urlencode({"page_number": 1})
+                while next_page_url:
+                    res = self.session.get(next_page_url)
+                    for item in res["result"]:
+                        yield item
+                    next_page_url = res.get("page", {}).get("next_page")
+
+        client = Client(url=base_url, principal=krb5_principal)
+
+        APIError = requests.exceptions.RequestException
+
     try:
         _add_to_cache(list(client.list_all_entities("users")))
-    except fasjson_client.errors.APIError as e:
+    except APIError as e:
         log.error("Something went wrong building cache with error: %s" % e)
         return {}
 
--- a/fedmsg_meta_fedora_infrastructure/fasshim.py
+++ b/fedmsg_meta_fedora_infrastructure/fasshim.py
@@ -1,13 +1,13 @@
 import collections
-import threading
+import logging
 import socket
 import string
+import threading
 from hashlib import sha256, md5
 
 _fas_cache = {}
 _fas_cache_lock = threading.Lock()
 
-import logging
 log = logging.getLogger("moksha.hub")
 
 try:
@@ -38,7 +38,9 @@ hardcoded_avatars = {
     # Ask tflink before actually putting this in place though.  we need
     # a nice small square version.  It'll look great!
     # In the meantime, we can use this temporary logo.
-    'taskotron': 'https://apps.fedoraproject.org/img/icons/taskotron-{size}.png'
+    'taskotron': (
+        'https://apps.fedoraproject.org/img/icons/taskotron-{size}.png'
+    )
 }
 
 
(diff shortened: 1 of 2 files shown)


fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -90,6 +90,50 @@ def avatar_url_from_email(email, size=64, default='retro', dns=False):
...
+    
+    try:
+        client = fasjson_client.Client(url=creds.get('base_url', default_url))
+    except fasjson_client.errors.ClientSetupError as e:
+        log.error("Something went wrong setting up fasjson client with error: %s" % e)
+    try:
+        response = client.list_users(page_size=1000)
+        _add_to_cache(response.result)
+        total_pages = response.page.get('total_pages')
+        for x in range(1, total_pages):
```
abompard, line 117: You're requesting page 1 twice, because that's what you got before the for loop. Also, `range()` will stop at the int before the end boundary so you'll miss the last page.
You can probably structure that around a while loop to make things simpler (increment the page number on each step until it reaches `total_pages`). Or maybe simplicity is in the eye of the beholder here ;-)

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -90,6 +90,50 @@ def avatar_url_from_email(email, size=64, default='retro', dns=False):
...
+
+def _add_to_cache(users):
+    global _fas_cache
+
+    for user in users:
+        nicks = user.get('ircnicks', [])
+        for nick in nicks:
+            _fas_cache[nick] = user['username']
+        
+        emails = user.get('emails')
```
abompard, line 133: Default to an empty list here too, otherwise the for loop may fail.

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -90,6 +90,50 @@ def avatar_url_from_email(email, size=64, default='retro', dns=False):
...
+    except fasjson_client.errors.ClientSetupError as e:
+        log.error("Something went wrong setting up fasjson client with error: %s" % e)
+    try:
+        response = client.list_users(page_size=1000)
+        _add_to_cache(response.result)
+        total_pages = response.page.get('total_pages')
+        for x in range(1, total_pages):
+            users = client.list_users(page_size=1000, page=x).result
+            _add_to_cache(users)
+        raise fasjson_client.errors.APIError
```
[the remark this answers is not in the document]
StephenCoady: No, that was accidentally left in from debugging, thanks :facepalm:

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -90,6 +90,50 @@ def avatar_url_from_email(email, size=64, default='retro', dns=False):
...
+    log.warn("No previous fasjson cache found. Looking to rebuild.")
+
+    try:
+        import fasjson_client
+    except ImportError:
+        log.warn("No fasjson-client installed.  Not caching fasjson.")
+    
+    creds = config['fas_credentials']
+    
+    default_url = 'https://admin.fedoraproject.org/accounts/'
```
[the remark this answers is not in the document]
StephenCoady: Yeah, I am leaving all the apps with the defaults they have currently so that it will be easier to find. Also, I am changing the deployment config for each one locally anyway.

fedmsg_meta_fedora_infrastructure/fasshim.py
StephenCoady: thanks, I thought I had set the for to start at `2`, obviously not. Yeah a while loop is simpler, will do.

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -110,16 +111,22 @@ def make_fasjson_cache(**config):
...
+        return {}
     try:
         response = client.list_users(page_size=1000)
         _add_to_cache(response.result)
         total_pages = response.page.get('total_pages')
-        for x in range(1, total_pages):
-            users = client.list_users(page_size=1000, page=x).result
+        # we want to start at the second page, since we already have the first
+        page_number = 2
+        while page_number < total_pages+1:
```
abompard, line 121: For the record I would have used something like:
```python
page_number = 0
next_page_exists = True
while next_page_exists:
    page_number += 1
    response = client.list_users(page_size=1000, page=page_number)
    _add_to_cache(response.result)
    next_page_exists = page_number < response.page["total_pages"]
```
This way the only visible API call is in the while loop, you don't do one just for the 1st page and then more, so if you have to change the page_size one day you only do it once and you can't forget the other place.

I'm a very strong believer in the "don't repeat yourself" principle, and it influences my writing a lot, but your way of doing it works too. It was just a FYI :-)

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -90,6 +90,50 @@ def avatar_url_from_email(email, size=64, default='retro', dns=False):
...
+    global _fas_cache
+    if _fas_cache:
+        return _fas_cache
+    
+    log.warn("No previous fasjson cache found. Looking to rebuild.")
+
+    try:
+        import fasjson_client
+    except ImportError:
+        log.warn("No fasjson-client installed.  Not caching fasjson.")
```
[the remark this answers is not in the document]
abompard: I'm still not sure whether it's better to raise an exception here or to disable the lookup. Let's leave that decision to the app owners.

fedmsg_meta_fedora_infrastructure/fasshim.py
StephenCoady: I added this because its better :)

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -113,15 +113,13 @@ def make_fasjson_cache(**config):
...
-        # we want to start at the second page, since we already have the first
-        page_number = 2
-        while page_number < total_pages+1:
-            users = client.list_users(page_size=1000, page=page_number).result
-            _add_to_cache(users)
-            page_number += 1
+        page_number = 0
+        next_page_exists = True
+        while next_page_exists:
+            page_number +=1
```
abompard, line 119: Flake8 is going to complain about a missing space right of the operator.

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -113,15 +113,13 @@ def make_fasjson_cache(**config):
...
+            response = client.list_users(page_size=1000, page=page_number)
+            _add_to_cache(response.result)
+            next_page_exists = page_number < response.page.get("total_pages")
```
abompard, line 122: Do you envision a case where `total_pages` will not be present in the page dict? If you know it's there you can just do `page["total_pages"]`. Using the `get` accessor is only useful if you want the call to return None if the key is absent (which would also crash the code here BTW, since comparing a number and None will raise an exception) or return a default value.

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -133,13 +131,20 @@ def _add_to_cache(users):
     global _fas_cache
 
     for user in users:
-        nicks = user.get('ircnicks', [])
-        for nick in nicks:
+        nick = user.get('ircnick')
```
abompard, line 134: I don't think a user has a `ircnick` property, it's always the multi-valued `ircnicks`.

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -133,13 +131,20 @@ def _add_to_cache(users):
...
+        if nick:
             _fas_cache[nick] = user['username']
-        
-        emails = user.get('emails', [])
-        for email in emails:
+        else:
+            nicks = user.get('ircnicks', [])
+            for nick in nicks:
+                _fas_cache[nick] = user['username']
+        email = user.get('email')
```
abompard, line 141: Same here I don't think the user model has an `email` property, it's all in `emails`.

fedmsg_meta_fedora_infrastructure/fasshim.py
StephenCoady: This is the original check that I just moved, its the FAS attribute `ircnick` not FASJSON's.

fedmsg_meta_fedora_infrastructure/fasshim.py
StephenCoady: Same as nick above.

fedmsg_meta_fedora_infrastructure/fasshim.py
StephenCoady: Good point, I'll make that change.

fedmsg_meta_fedora_infrastructure/fasshim.py
abompard: Ah I see. There's two ways of going at it, you can either deal with the complexity of the 2 sources here, or you can modify the results of the source that is going to be deprecated (FAS) to make it compatible with the new one (adding `ircnicks` and `emails` values).
I tend to favor the latter option because then when we remove the code of the deprecated option, we just delete the block and there's no specific handling anywhere else in the codebase that we must remember to remove too.

StephenCoady: ok, so really the way it was _before_ I created the add_to_cache method was better because it reduces coupling. If I revert that section I don't need to make anything compatible and we can just remove the fas sections once it is deprecated.

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -90,6 +90,49 @@ def avatar_url_from_email(email, size=64, default='retro', dns=False):
...
+    try:
+        client = fasjson_client.Client(url=creds.get('base_url', default_url))
+    except fasjson_client.errors.ClientSetupError as e:
+        log.error("Something went wrong setting up fasjson client with error: %s" % e)
+        return {}
+    try:
+        _add_to_cache(list(client.list_all_entities("users")))
+    except fasjson_client.errors.APIError as e:
+        log.error("Something went wrong building cache with error: %s" % e)
+        continue
```
abompard, line 119: When I run the tests with tox I get `SyntaxError: 'continue' not properly in loop`. It should be `return {}`.

StephenCoady: thanks, not sure how that slipped through. fixed now.

StephenCoady:
OK so it looks like the python 2 tests are failing on `develop` at the moment, so I am ignoring them. I am seeing failures on `py36-six17` also, but not on `py36-six19` for some reason. Still investigating but I'm not 100% sure what is happening.

fedmsg_meta_fedora_infrastructure/fasshim.py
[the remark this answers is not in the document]
pypingou: I'd say let's raise an exception, the users have specified in their configuration they want to use fasjson, so let's tell them if they cannot.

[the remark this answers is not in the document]
nphilipp: This looks like it's resolved. I'll attempt to rebase the PR.

nphilipp:
The latest push adds a minimal shim to stand in for `fasjson-client` which only works on Python >= 3.6.

nphilipp:
Looks like there are some deps that need to be pinned to a version that works with Python 2.7.

nphilipp:
If CI is happy with this, we should probably squash the changes.

nphilipp:
Hmm, for some reason with `requests-gssapi`, the CI didn't find `requests`. Replacing the dash with an underscore worked around it locally.

nphilipp:
So, the latest push reverses the order of `requests` and `requests-gssapi` in `install_requires` and now CI is happy on Python 2.7, too.

fedmsg_meta_fedora_infrastructure/fasshim.py
```diff
@@ -1,13 +1,14 @@
 import collections
+import os
 import threading
 import socket
 import string
+import logging
```
pypingou, line 6: should we order them alphabetically? At least for the two we're adding/moving?

pypingou (commented):
One stupid comment, otherwise looks quite nice to me :+1: for me

StephenCoady (commented):
I can't approve this PR as its originally mine, but the new changes look good to me @nphilipp.

StephenCoady:
@nphilipp do you want to approve and merge this? see my comment above

fedmsg_meta_fedora_infrastructure/fasshim.py
nphilipp: 👍

PR #504, opened 2021-03-19
Add information about the namespace in the short description of git messages

pypingou:
Currently the short description was talking about the git push to a repo
but did not include the namespace of the repo. This commit fixes that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ install:
  - python setup.py install
  - pip install arrow
 
-script: python setup.py test
+script: python setup.py test -q
 
 env:
   global:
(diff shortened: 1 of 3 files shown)


AdamSaleh:
My only question is, do we still want to even have python 2.7 there? I would assume we want to migrate/force people to migrate asap. Does this thing run elsewhere than on our infra anyway?

pypingou:
> My only question is, do we still want to even have python 2.7 there?

I kept it for RHEL/CentOS 7 but I honestly no longer care if it fails :)

> I would assume we want to migrate/force people to migrate asap. Does this thing run elsewhere than on our infra anyway?

Yes, I know of a couple of people who use/used it outside of our infra.

AdamSaleh (approved):
Looks good.

pypingou:
Thanks for the review :)

PR #505, opened 2021-03-23
fasjson: Remove version from default base URL

nphilipp:
The fasjson client adds this by itself, so the shim for it should do
likewise.

Signed-off-by: Nils Philippsen <nils@redhat.com>


StephenCoady (approved):
lgtm, thanks!

PR #508, opened 2021-11-04
waiverdb: handle newer messages

AdamWill:
waiverdb has changed its message format three times since the
processor was written, but it has never been updated. This should
record and handle all four types of message.


AdamWill:
Note - I'm using CI to run the tests here, haven't run them locally, could still have errors.

AdamWill:
welp, apparently there is no CI. ran tox locally and it didn't show any errors in these tests (hit six errors and a failure in some other tests).

